Issue #34, opened 2023-06-28
Should we use "Subject" and "Resource" rather than "Principal" and "Asset" ?


tulshi:
### Discussed in https://github.com/SGNL-ai/authzapi/discussions/22

<div type='discussions-op-text'>

<sup>Originally posted by **baboulebou** June 28, 2023</sup>
"Subject" and "Resource" have been used in the industry since XACML, should we just stick to those terms?</div>

tulshi:
I have a minor concern about changing "asset" to "resource": In my mind a "resource" indicates a passive entity such as a database, whereas an "asset" can include active entities such as methods or services. So I have a slight bias toward asset over resource.

I have a similar, if not stronger concern about the word "subject" (over "principal"), in that a subject to me feels like a passive entity and not an initiator of the request.

This might just be due to my not being active in any XACML discussions, but I would like us to choose terms that we all think are best suited going forward and not necessarily be bound by XACML, especially because it is not very popular in the industry today.

baboulebou:
Interesting, In my own mind, a "subject" is an active actor that needs to be authorized (human or not), and a "Resource" is anything which access needs to be protected, not only DBs: could be IoT devices too for example.  So exactly the semantics you have for "principal" and "asset"... 

Actually, just had a look at the xacml spec, and even there the meanings of the terms are not constrained to Humans and DBs, Subjects are actors:

"Resource : Data, service or system component"
"Subject: An actor whose attributes may be referenced by a predicate"

Anyway, not really defending xacml here (sorry @davidjbrossard ), but at least these terms provide some continuity and carry the same semantics.

tulshi:
Happy to go with the proposed change from "Principal" to "Subject" and from "Asset" to "Resource". Let's see if anyone else has any comments on this topic.

davidjbrossard:
In XACML, as Alex mentioned, we have the concept of categories. They
typically map to grammatical functions. XACML also lets you define custom
categories (but I've seldom seen anyone do so). The default categories in
XACML are subject, resource, action, and environment. They actually have
longer urn-style names and subject actually has different kinds (recipient,
access, etc...)

Lately the team at Axiomatics has been thinking that adding a category is
superfluous. OPA doesn't define categories at all. All we really need is
just key-value pairs e.g. acme.user.name and acme.action. We also noticed
that using namespace-style naming conventions is really helpful.

davidjbrossard:
[...]

PR #36, opened 2023-06-29
Added reverse Search section: search Principals allowed to access a g…

baboulebou:
…iven Asset.


authorization-api-1_0.md
```diff
@@ -599,7 +602,184 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
 ~~~
 {: #example-search-response title="Example of an Search Response"}
 
+## Reverse-Search API
+The Access Reverse Search API does the reverse of the Search API: it enables a PEP or client to find out all the principals that can access a given asset.
```
tulshi, line 606: Instead of calling this "reverse search", should we use a terminology like "principal search" (in contrast with "asset search")?

tulshi: Responded in the issue #34

baboulebou: Changed to "Subject search".

authorization-api-1_0.md
```diff
@@ -599,7 +602,184 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
 ~~~
 {: #example-search-response title="Example of an Search Response"}
 
+## Subject Search API
+The Access Subject Search API does the reverse of the Search API: it enables a PEP or client to find out all the principals that can access a given asset.
```
tulshi, line 606: Do you want to rename the existing "Search API" to be "Resource Search API" (and this one is the "Subject Search API")?

baboulebou: Done in latest commit.

authorization-api-1_0.md
```diff
@@ -597,9 +600,186 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
...
+
+pageToken:
+: OPTIONAL. A string value that is returned in a previous Subject Search Response ({{subject-search-response}}), which indicates that the request is a continuation of a previous request
+
+pageSize:
+: OPTIONAL. The maximum number of `decision` items in a Subject Search Response ({{subject-search-response}}). The API MAY return a smaller number of items but SHOULD NOT return a number of items that is greater than this value
+
+The content of a Subject Search Request body is a JSON object with the following fields:
+
+asset:
```
tulshi, line 621: resource?

authorization-api-1_0.md
```diff
@@ -597,9 +600,186 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
...
+
+The following is a non-normative example of a Subject Search Request:
+
+~~~ http
+POST /subjectsearch HTTP/1.1
+Host: pdp.mycompany.com?pageToken="NWU0OGFiZTItNjI1My00NTQ5LWEzYTctNWQ1YmE1MmVmM2Q4"&pageSize=2
+Authorization: <myoauthtoken>
+
+{
+  "asset": {
```
tulshi, line 635: "resource"

authorization-api-1_0.md
```diff
@@ -597,9 +600,186 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
...
+~~~
+{: #example-subject-search-request title="Example Subject Search Request"}
+
+### Subject Search Response {#subject-search-response}
+The success response to a Subject Search Request is a Subject Search Response. It is a HTTP response of type `application/json`. Its body is a JSON object that contains the following fields:
+
+iat:
+: REQUIRED. The issued at time in `integer` format, expressed as epoch milliseconds
+
+principal:
```
tulshi, line 655: subject

authorization-api-1_0.md
```diff
@@ -597,9 +600,186 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
...
+: REQUIRED. The principal for which the response is being issued. The format of this field is as described in the Assets section ({{principals}})
```
tulshi, line 656: The subject for which... Also fix reference to {{subjects}}

tulshi (commented):
Please update old references to asset and principals to resource and subject respectively

authorization-api-1_0.md
baboulebou: All remaining typos/renames should be done now.

authorization-api-1_0.md
```diff
@@ -597,9 +600,186 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
...
+    "actions": ["delete", "read", "write"],
+    "attributeNames": [
+        "author",
+        "createDate",
+        "lastUpdated"
+    ],
+    "subject": {
+        "id": "alex@3edges.com"
+    },
+    "reason_ids": [0,2,3]
```
tulshi, line 696: You probably need a "decision" string here that has either an "allow" or a "deny" value

authorization-api-1_0.md
```diff
@@ -597,9 +600,186 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
...
+
+reasons:
+: OPTIONAL. An array of Reason Objects ({{reason-object}}) that describe the reason for each reason identified in the `decisions` field. This field is REQUIRED if there is at least one decision in the `decisions` field that specifies a `reason_ids` field. The content of the `reasons` field MUST provide details of every identifier in the `reason_ids` fields in the `decisions` array.
+
+nextPageToken:
+: OPTIONAL. A string that MAY be used in a Search Request to fetch the next set of responses.
+
+#### Subject Query Result {#subject-query-result}
+A Subject Query Result is JSON object combining a subject, a list of resource attribute names and an action. Given that a Subject Query result is expected to be the response to a Subject Search, only positive matches should be returned; i.e., only those subjects that match the search criteria (those subjects that are allowed to access the provided Resource Attributes). Any Subjects absent from the results do not have any access to the Resource. 
+A Subject Query Result has the following fields:
```
tulshi, line 669: The example you write below is correct (except it should have the "decision" string in each object), but the description here needs to be updated to say that the response is an array of decisions. We should also call this top-level array something other than "decisions", since we are using that in the low-level decision key.

authorization-api-1_0.md
baboulebou: Actually, in the section above this, I stated: 
"Given that a Subject Query result is expected to be the response to a Subject Search, only positive matches should be returned; i.e., only those subjects that match the search criteria (those subjects that are allowed to access the provided Resource Attributes). Any Subjects absent from the results do not have any access to the Resource. "

If we also return the "Deny" subject, then we basically returning everybody, which is too much.

authorization-api-1_0.md
baboulebou: See comment above for the 1st part: again, I think we should only return positive results, we're unnecessarily increasing the size of the response, and the computation required to create the response imho. Anything not returned should be assumed denied.

authorization-api-1_0.md
tulshi: Yeah, I agree we should return positive results only. I wonder if we should still have the reason codes then, because they may be useful to clients only if the decision is "deny". I can see some value to it, but I feel it will be OK to not return the reasons.

baboulebou: Opened an issue about this, then closed it seeing this :) .

Anyway yes I tend to agree... ok will remove those from the results

authorization-api-1_0.md
tulshi: Agreed with your previous comment. I feel we can drop the "reasons" from the Subject Search Query Response too.

tulshi (approved):

Issue #46, opened 2023-11-06
revise use of "ipAddress" for subjects


tr33:
IP addresses belong to the network layer and may not be available to the PEP or may not affect many AuthZ decision requests.
If the subject is identified through its IP address, then it should be the subjects "id" attribute. 

Also, IPv6 addressing schemes can be rather complex and error-prone, due to multiple variations in syntax:
eg. those are different variants of THE SAME ip address:

- fe80::AbCD:3:2:0001
- Fe80:0:0:0:abcd:0003:0002:0001
- Fe80:0:0:0:abcd:0003:0002:0001%eth1

Its horrid to implement a unification/syntax check on PDPs/PEP side.

recommendation:
- remove the "ipAddress" attribute until further specified.
- or define a dedicated structure with optional, freely defined attributes where an "ipAddress" can be defined as a custom attribute without further specification.

baboulebou:
My only concerns about these Subject attributes is that they are geared towards human Subjects. A smart IoT device or a Web Service could also be Subjects. Seems to me that Subject could be a JSON object with whatever attributes make sense to the implementer. Just enforce an `id` and leave the rest open...

davidjbrossard:
See #47

PR #57, opened 2023-12-11
incorporated subject identifiers from RFC9493

tulshi:


tr33:
question to definition of 
> "format": "ip_address",

the examples only contain IPv4 adresses. But IPv4 is overaged and new systems must be IPv6 compatible.
however, both address formats will remain in the wild for sure a long time.

Question: should the **ip address be typed** in order to distinguish between multiple versions an formats?
like to distinguish between address netmask (a.b.c.d/24) and  IPv6?

also, IPv6 address can be formatted ambiguously (see [Issue #46](https://github.com/openid/authzen/issues/46) )

ogazitt:
I get the value of relying on another established spec for subject identification, but I think there are 2-3 important use-cases missing:
1. `"format": "jwt"`: a base64 encoded JWT
2.`"format": "sub"`: a subject (identity) which is the `sub` claim from a structured token like a JWT
3. `"format": "string"`: an opaque string (the PEP passes a string that is meaningful in the context of the policy being evaluated - e.g. it's a key that can be looked up by the policy and resolved into a subject)

In addition, would it make sense to allow something like `"format": "anonymous"` to be able to enforce policy that isn't subject-specific? For example, time of day or IP address range enforcement?

api/authorization-api-1_0.md
```diff
@@ -46,6 +46,7 @@ normative:
   RFC6750: #OAuth 2.0 Bearer Tokens
   RFC8259: #JSON
   RFC9110: # HTTP Semantics
+  RFC9493: # Subject Identifiers for SETs
```
adeinega, line 49:
```suggestion
  RFC9493: # Subject Identifiers for Security Event Tokens
```

api/authorization-api-1_0.md
```diff
@@ -128,27 +129,61 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305716
...
+~~~
+{: #ipaddresssubjectexample title="Example IP Address format Subject Identifier"}
 
-deviceId:
-: OPTIONAL. A field, whose value is of type `string`, which uniquely identifies the device of the Subject
+### Device Identifier {#subject-device-id}
+The Device Identifier of the Subject. The Device Identifier Subject Identifier Format is identified by the name `device_id`. It identifies the Device Identifier of the subject, identified with a `device_id` field, whose value is a string representation of the device identifier.
 
-The following non-normative example describes a Subject:
+The following is a non-normative example of a Subject Identifier Format of type  Device Identifier:
```
adeinega, line 152:
```suggestion
The following is a non-normative example of a Subject Identifier Format of type Device Identifier:
```

api/authorization-api-1_0.md
```diff
@@ -383,7 +418,8 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305716
 
 {
   "subject": {
-    "id": "atul@sgnl.ai",
+    "format": "email",
```
adeinega, line 421: I would suggest using `account`, `iss_sub` or `opaque` in examples rather than `email`.

`email` is a problematic subject identifier because canonicalization algorithms for email addressees aren't that well defined. In some edge cases, it's barely possible to canonicalize email addresses, as a bright example, both user@gmail.com and user+label1@gmail.com "represent" the same email address.

api/authorization-api-1_0.md
```diff
@@ -751,8 +815,21 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305720
 
 # IANA Considerations {#IANA}
 
-TBS
+The following Subject Identifier Formats SHALL be added to the "Security Event Identifier Formats" Registry as described in "Subject Identifiers for Security Event Tokens" {{RFC9493}}:
+
+## IP Address {#ipaddress-registry-entry}
+
+* Format Name: ip_address
+* Format Description: A value that describes a subject through its IP Address
```
adeinega, line 823:
```suggestion
* Format Description: A value that describes a subject through its IP address
```

davidjbrossard:
I am closing this PR as we've since done major rework to the API that voids the work done in this PR.

PR #105, opened 2024-05-20
Add support for test output formats

alexolivier:
Output to console:

```shell
yarn test https://authorizer.domain.com console
```

Output as markdown:

```shell
yarn test https://authorizer.domain.com markdown
```


interop/authzen-todo-backend/test/runner.ts
```diff
@@ -0,0 +1,98 @@
...
+          Authorization: AUTHZEN_PDP_API_KEY,
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(REQ),
+      });
+
+      const data = await response.json();
+      const RSP = data.decision || false;
+
+      results.push({
```
ogazitt, line 39: Thanks @alexolivier ! Sorry for being so late with this review, just got a chance to look at it.

My only feedback is that for the `console` output, I think we should emit the result immediately - which will preserve the current behavior of giving the user immediate feedback as the tests are executed. 

For the `markdown` output, I think the current code makes perfect sense.

ogazitt (changes requested):

interop/authzen-todo-backend/test/runner.ts
alexolivier: Pushed an update to log results immediately when using console mode

ogazitt (approved):

PR #113, opened 2024-06-12
refactor API spec to remove search and boxcarred apis and cleanup

ogazitt:
* removed Evaluations and Search APIs
* cleaned up all JSON
* added subject IDs as per RFC9493 (subsuming @tulshi's earlier PR)
* added required `type` field for subjects and objects as discussed
* added required `name` field for actions as discussed
* cleanly factored HTTPS binding into the transports section
* added example annotations for every JSON example
* fixed all the bullets that weren't rendering correctly
* grammar and typos

### DIFF
--- a/api/authorization-api-1_0.md
+++ b/api/authorization-api-1_0.md
@@ -19,6 +19,7 @@ kw:
   - Cedar
   - PDP
   - PEP
+  - ALFA
 # date: 2022-02-02 -- date is filled in automatically by xml2rfc if not given
 author:
 - role: editor # remove if not true
@@ -103,9 +104,32 @@ A Subject is a JSON ({{RFC8259}}) object that contains any number of key-value p
 `id`:
 : REQUIRED. The unique identifier of the Subject, scoped to the `type`, within the scope of the PEP. If specified, the value MUST follow the format specified by the `Subject Identifiers for Security Event Tokens` specification {{RFC9493}}.
 
-A Subject MUST minimally contain a `type` and an `id`, and MAY contain any number of additional fields.
+A Subject MUST minimally contain a `type` and an `id`, and MAY contain zero or more additional fields.
 
 The following is a non-normative example of a subject:
+
+~~~ json
+{
+  "type": "user",
+  "id": "alice@acmecorp.com"
+}
+~~~
+{: #subject-example title="Example Subject"}
+
+#### Subject Identifier {#subject-identifier}
+The `id` field of a Subject MAY be any valid JSON value. It MAY be a string, or it MAY be a structured identifier. For example, it MAY follow the format specified by the `Subject Identifiers for Security Event Tokens` specification {{RFC9493}}.
+
+The following is a non-normative example of a subject identifier as a simple string:
+
+~~~ json
+{
+  "type": "user",
+  "id": "alice@acmecorp.com"
+}
+~~~
+{: #subject-identifier-example-simple title="Example of Simple Subject Identifier"}
+
+The following is a non-normative example of a Subject Identifier in the {{RFC9493}} Email Identifier Format:
 
 ~~~ json
 {
@@ -116,39 +140,36 @@ The following is a non-normative example of a subject:
   }
 }
 ~~~
-{: #subject-example title="Example Subject"}
-
-#### Subject Identifier {#subject-identifier}
-The `id` field of a Subject MUST follow the format specified by the `Subject Identifiers for Security Event Tokens` specification {{RFC9493}}. A Subject MUST include at least one identifier, but MAY include more than one identifier if the format of the field is of type `aliases`.
-
-The following is a non-normative example of a Subject Identifier in the Email Identifier Format:
-
-~~~ json
-{
-  "format" : "email",
-  "email": "alice@acmecorp.com"
-}
-~~~
-{: #subject-identifier-example title="Example Subject Identifier"}
+{: #subject-identifier-example-rfc9493 title="Example Subject Identifier as RFC9493 Subject"}
 
 #### Subject Type {#subject-type}
 Since {{RFC9493}} only concerns itself with the *format* of the identifier and not its *type*, every Subject MUST also include a string-valued `type` field, which identifies the type of Subject.
 
-The following is a non-normative example of a Subject of type `group` with a Subject Identifier in the Email Identifier Format:
+The following is a non-normative example of a Subject of type `group` with a Subject Identifier as a simple string:
+
+~~~ json
+{
+  "type": "group",
+  "id": "engineering@acmecorp.com"
+}
+~~~
+{: #subject-type-group-example title="Example Group Subject Type"}
+
+The following is a non-normative example of a Subject of type `group` with a Subject Identifier in the {{RFC9493}} Email Identifier Format:
 
 ~~~ json
 {
   "type": "group",
   "id": {
     "format" : "email",
-    "email": "alice@acmecorp.com"
+    "email": "engineering@acmecorp.com"
   }
 }
 ~~~
-{: #subject-type-example title="Example Subject Type"}
+{: #subject-type-example-rfc-9493 title="Example Subject Type in RFC9493 Format"}
 
 #### Subject Attributes {#subject-attributes}
-Many authorization systems are stateless, and expect the client (PEP) to pass in any attributes that are expected to be used in the evaluation of the authorization policy. To satisfy this requirement, Subjects MAY include any number of additional attributes as key-value pairs.
+Many authorization systems are stateless, and expect the client (PEP) to pass in any attributes that are expected to be used in the evaluation of the authorization policy. To satisfy this requirement, Subjects MAY include zero or more additional attributes as key-value pairs.
 
 An attribute can be single-valued or multi-valued. It can be a primitive type (string, boolean, number) or a complex type such as a JSON object or JSON array.
 
@@ -157,10 +178,7 @@ The following is a non-normative example of a Subject which adds a string-valued
 ~~~ json
 {
   "type": "user",
-  "id": {
-    "format" : "email",
-    "email": "alice@acmecorp.com"
-  },
+  "id": "alice@acmecorp.com",
   "department": "Sales"
 }
 ~~~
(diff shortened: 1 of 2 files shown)


api/authorization-api-1_0.md
```diff
@@ -15,6 +15,7 @@ kw:
   - Access Management
   - XACML
```
davidjbrossard, line 16: If we're mentioning XACML and Cedar, we should mention ALFA

api/authorization-api-1_0.md
```diff
@@ -67,187 +67,292 @@ normative:
...
+
+- Can Alice view document #123?
+- Can Alice view document #123 at 16:30 on Tuesday, June 11, 2024?
+- Can a manager print?
 
 ## API Version
 This document describes the API version 1. Any updates to this API through subsequent revisions of this document or other documents MAY augment this API, but MUST NOT modify the API described here. Augmentation MAY include additional API methods or additional parameters to existing API methods, additional authorization mechanisms, or additional optional headers in API requests. All API methods for version 1 MUST be immediately preceded by the relative URL path `/v1/`.
 
 ## Information Model
+The information model for requests and responses include the following entities: Subject, Action, Resource, Context, and Decision. These are all defined below.
```
davidjbrossard, line 93: So, I'm now having second thoughts about allowing additional categories... For instance, what if I want to ask: "Can Alice send $100 to Bob?". Historically I modeled this as Alice in Subject, send in Action and the rest (amount and recipient) in resource. But... we could also introduce a recipient-subject category. Too much?

api/authorization-api-1_0.md
```diff
@@ -67,187 +67,292 @@ normative:
...
+    "format" : "email",
+    "email": "alice@acmecorp.com"
+  }
+}
 ~~~
+{: #subject-example title="Example Subject"}
 
-An attribute can be single-valued or multi-valued. It can be a primitive type (string, boolean...) or a complex type such as a JSON object.
+#### Subject Identifier {#subject-identifier}
+The `id` field of a Subject MUST follow the format specified by the `Subject Identifiers for Security Event Tokens` specification {{RFC9493}}. A Subject MUST include at least one identifier, but MAY include more than one identifier if the format of the field is of type `aliases`.
```
davidjbrossard, line 122: What I don't like with RFC 9493 is that it now forces us to specify format. It makes creating an AuthZEN request harder on the developer. It goes back to the same verbose nature of a XACML request where you had to specify datatype.

api/authorization-api-1_0.md
```diff
@@ -67,187 +67,292 @@ normative:
...
+{
+  "name": "can_read"
+}
+~~~
+{: #action-example title="Example Action"}
+
+#### Common Action Values
+Since many services follow a Create-Read-Update-Delete convention, a set of common Actions are defined. That said, an Action may be custom, which could be specific to the application being accessed or shared across applications but not listed in the common Actions below.
+
+The following common Actions are defined:
```
davidjbrossard, line 256: I think the values are not "ABAC" enough. Why specify values in the first place? And why not just use the verbs e.g. view, edit, update, access rather than an artificial can_read? Also, why use underscore?

api/authorization-api-1_0.md
```diff
@@ -260,22 +365,23 @@ A Reason Field is a JSON object that has keys and values of type `string`. The f
...
 ~~~
+{: #reason-example title="Example Reason"}
 
 ###### Reason Object {#reason-object}
 A Reason Object specifies a particular reason. It is a JSON object that has the following fields:
 
-id:
-: REQUIRED. A numeric value of that specifies the reason within the scope of a particular response
+`id`:
+: REQUIRED. A numeric value of that specifies the reason within the scope of a particular response.
```
davidjbrossard, line 374: Why make it a numeric value?

api/authorization-api-1_0.md
```diff
@@ -260,22 +365,23 @@ A Reason Field is a JSON object that has keys and values of type `string`. The f
...
 
-reason_admin:
+`reason_admin`:
```
davidjbrossard, line 376: Rather than having different fields for different audiences, why not have a reason field and then an audience field?

api/authorization-api-1_0.md
```diff
@@ -432,40 +434,33 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305716
...
 {
   "subject": {
-    "id": "omri@aserto.com",
-  },
-  "action": {
-    "action": "can_read"
+    "type": "user",
+    "id": {
+      "format": "email",
+      "email": "omri@aserto.com"
```
davidjbrossard, line 440: Should we use alice@example.com?

api/authorization-api-1_0.md
```diff
@@ -476,449 +471,74 @@ X-Request-ID: bfe9eb29-ab87-4ca3-be83-a1d5d8305716
...
-  "nextPageToken": "1DlR0Em5panAPy5llasLPfNUpDztEKgTDKF2I5gPwymnc"
-}
-~~~
-{: #example-subject-search-response title="Example of a Subject Search Response"}
-
-# Transport
+The following errors are indicated by the status codes defined below:
 
-## REST Binding
+| Code | Description  | HTTPS Body Content |
```
davidjbrossard, line 481: We had a chat a while back that some proxies/gateways cannot inspect the response payload and rely on HTTP codes to determine access. Do we want to have a flag that allows the PDP to return HTTP 403 in case of decision false?

davidjbrossard (commented):
I added other comments that I forgot to explicitly add to this review.

api/authorization-api-1_0.md
ogazitt: I don't see a reason to revisit SARE / PARC...

There are a few ways of breaking up the action and resource, but I think the most natural way to do it is that the Action is "Send" and the Resource is of type:user, id:Bob, amount:$100.

api/authorization-api-1_0.md
ogazitt: I struggled with this too, but I thought we all agreed to use RFC 9493 because it has some pre-defined formats.

api/authorization-api-1_0.md
ogazitt: These are meant to read as permissions.  The convention that ReBAC systems use to differentiate permissions from relations are that permissions start with `can_` (and are not directly assignable).

I personally don't think the "common Actions" are that useful, but they were in the current draft and I believe there was some consensus that we wanted common values.

api/authorization-api-1_0.md
ogazitt: Leftover from Atul's spec. I am happy with this being a string actually.

api/authorization-api-1_0.md
ogazitt: sure

api/authorization-api-1_0.md
ogazitt: I agree that if we add this, we would want to control it with a flag.

api/authorization-api-1_0.md
ogazitt: done

api/authorization-api-1_0.md
ogazitt: resolved during discussion - keeping SARC model

api/authorization-api-1_0.md
ogazitt: Discussed during call - make RFC 9493 OPTIONAL

api/authorization-api-1_0.md
ogazitt: leaving this in the spec as-is for now, since it doesn't appear to be blocking the merge of the PR and getting us into a better state. we can discuss the reason fields as a separate PR.

api/authorization-api-1_0.md
ogazitt: done

api/authorization-api-1_0.md
ogazitt: discussed during the call - agreed to leave this out of scope for now (and certainly for this PR).

api/authorization-api-1_0.md
```diff
@@ -67,187 +67,292 @@ normative:
...
+    "email": "alice@acmecorp.com"
+  }
+}
+~~~
+{: #subject-type-example title="Example Subject Type"}
 
-deviceId:
-: OPTIONAL. A field, whose value is of type `string`, which uniquely identifies the device of the Subject
+#### Subject Attributes {#subject-attributes}
+Many authorization systems are stateless, and expect the client (PEP) to pass in any attributes that are expected to be used in the evaluation of the authorization policy. To satisfy this requirement, Subjects MAY include any number of additional attributes as key-value pairs.
```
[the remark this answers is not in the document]
ogazitt: incorporated

api/authorization-api-1_0.md
ogazitt: discussed on the call, and changed to string

api/authorization-api-1_0.md
```diff
@@ -67,187 +67,292 @@ normative:
...
 
 --- middle
 
 # Introduction
-
 Computational services often implement access control within their components by separating Policy Decision Points (PDPs) from Policy Enforcement Points (PEPs). PDPs and PEPs are defined in XACML ({{XACML}}) and NIST's ABAC SP 800-162. Communication between PDPs and PEPs follows similar patterns across different software and services that require or provide authorization information. The Authorization API described in this document enables different providers to offer PDP and PEP capabilities without having to bind themselves to one particular implementation of a PDP or PEP.
 
 ## Model
-The Authorization API is a transport-agnostic API published by the PDP, to which the PEP acts as a client. The Transport section of this specification will detail possible bindings e.g. REST. 
+The Authorization API is a transport-agnostic API published by the PDP, to which the PEP acts as a client. The Transport section of this specification details possible bindings, such as HTTPS or gRPC.
```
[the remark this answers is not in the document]
ogazitt: Discussed on the call, agreed to leave the HTTPS binding as-is for now, and add a section 2.2 if/when we get a gRPC action. Action item - have Gert and Alex discuss whether the JSON payload documentation is reasonable to be able to express as a gRPC protobuf contract.

tulshi (approved):

davidjbrossard (approved):

PR #168, opened 2024-10-09
Plainid 1.1-preview

vladiber:


ogazitt:
Thanks @vladiber, I verified that your PDP passes the 1.0-implementers-draft and 1.1-preview test suites.

It doesn't seem to pass the 1.0-preview test suite anymore. Is this by design or an accident? I don't think it matters in the long run, because once all the implementations are 1.0-implementers-draft compliant, I will remove that option from the pdps.json file and the test suite.

The yarn.lock file has a huge changeset, which shouldn't be necessary for this PR. So I will instead open a new PR that makes the right changes and close this one.

ogazitt:
closing as per https://github.com/openid/authzen/pull/172

PR #170, opened 2024-10-11
fixed typo in action name for boxcarred evaluations

tulshi:


ogazitt (approved):
Good catch, thanks! Also updated the hackmd source document, https://hackmd.io/gNZBRoTfRgWh_PNM0y2wDA?view

PR #172, opened 2024-10-12
added PlainID to PDP list and interop results

ogazitt:


ogazitt:
This replaces PR #168

Issue #190, opened 2024-12-29
The interop backend uses the wrong resource IDs for `can_update_todo` and `can_delete_todo`


aaguiarz:
It's using the OwnerID, which is the user's email, instead of the Todo item ID

https://github.com/openid/authzen/blob/dcbac20955506c0601325d2d66bc8bebec49b1a4/interop/authzen-todo-backend/src/auth.ts#L56

https://github.com/openid/authzen/blob/dcbac20955506c0601325d2d66bc8bebec49b1a4/interop/authzen-todo-backend/src/auth.ts#L66

PR #191, opened 2024-12-29
Use proper ID for todo Item in can_update_todo and can_delete_todo

aaguiarz:
Fixes https://github.com/openid/authzen/issues/190


davidjbrossard (commented):
This looks good to me. @ogazitt thoughts?

ogazitt (approved):
Good catch!

PR #200, opened 2025-01-29
Refactor demo react app

alexolivier:
As ground work for introducing the next interop scenario, this PR updates the React application to use the more modern way to build react apps (Vite, Tanstack Query, Contexts etc).

CRA has long been deprecated and this resolves several security issues.

~~@ogazitt I'm not sure how you are deploying this as there isn't a Dockerfile in here, unless its done statically inwhich case `yarn build` will generate the assets in the `dist` folder.~~

Updated it to use the same `build` output folder to keep the existing behavior


ogazitt:
> As ground work for introducing the next interop scenario, this PR updates the React application to use the more modern way to build react apps (Vite, Tanstack Query, Contexts etc).
> 
> CRA has long been deprecated and this resolves several security issues.
> 
> ~@ogazitt I'm not sure how you are deploying this as there isn't a Dockerfile in here, unless its done statically inwhich case `yarn build` will generate the assets in the `dist` folder.~
> 
> Updated it to use the same `build` output folder to keep the existing behavior

Thanks @alexolivier! I'll review when I get back. One quick point - the current react app gets deployed by Netlify into a static bundle served from their CDN.

ogazitt (approved):

PR #225, opened 2025-02-25
Gw ajustments

vladiber:


interop/authzen-interop-website/package-lock.json
ogazitt, line 1: We already have a yarn.lock, so we don't want a package-lock.json. Can you please remove from the PR?

ogazitt (changes requested):

interop/authzen-interop-website/package-lock.json
vladiber: Done

ogazitt (approved):

PR #236, opened 2025-03-03
Fix Misalignment Between Request Payload and Description in PUT Todo Section

Thumimku:
I'm developing the **PDP evaluation endpoint** for **WSO2 Identity Server**, using the specification while also validating it against **Interop Scenarios**.

I noticed an inconsistency in the **[PUT todo section](https://authzen-interop.net/docs/scenarios/todo-1.1/#put-todosid)**—the descriptions do not align with the corresponding request and response payloads.

For example, the request payload description states:
> *The `resource.properties` contains an attribute called `ownerID`, which holds the owner's ID (as defined in the "Attributes" section above, where it is the owner's email address).*

However, the actual payload **does not contain a `properties` section**, making the description misleading.

To ensure accuracy and consistency, I am submitting this **PR to correct the documentation**.

### DIFF
--- a/interop/authzen-interop-website/docs/scenarios/todo-1.1/index.md
+++ b/interop/authzen-interop-website/docs/scenarios/todo-1.1/index.md
@@ -361,9 +361,12 @@ The `resource.properties` contains an attribute called `ownerID` which contains 
     "name": "can_update_todo"
   },
   "resource": {
-    "ownerID": "<email_of_owner>",
-    "type": "todo"
-  },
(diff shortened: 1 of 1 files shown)


ogazitt (approved):
Good catch!

ogazitt:
The "source of truth" payload document [here](https://hackmd.io/gNZBRoTfRgWh_PNM0y2wDA?view#PUT-todosid) has the correct payload, but the checked in file that this PR identifies was not properly updated.

PR #238, opened 2025-03-08
Add OpenFGA support and interop results

aaguiarz:
- Added support for stateful AuthZ services on the todo-application. It provides a way to specify how the state is written to each specific service.

- Added OpenFGA PDP info and interop results


interop/authzen-todo-backend/src/pdps.json
```diff
@@ -9,7 +9,8 @@
       "Open Policy Agent": "https://authzen-opa-proxy-preview.demo.aserto.com",
       "Permit.io": "https://permit-authzen-interop-1-0-preview.up.railway.app",
       "Thales AuthZForce": "https://restful-pdp-vji43cydea-uc.a.run.app",
-      "Topaz": "https://authzen-topaz-proxy-preview.demo.aserto.com"
+      "Topaz": "https://authzen-topaz-proxy-preview.demo.aserto.com",
+      "OpenFGA": "https://authzen-interop.openfga.dev/stores/01JG9JGS4W0950VN17G8NNAH3C"
```
ogazitt, line 13: Let's sort these alphabetically - so right after Open Policy Agent

interop/authzen-todo-backend/src/pdps.json
```diff
@@ -40,11 +42,12 @@
...
       "PlainID": "https://authzeninteropt.se-plainid.com",
       "SGNL": "https://authzen.sgnlapis.cloud",
-      "Topaz": "https://authzen-topaz-proxy.demo.aserto.com"
+      "Topaz": "https://authzen-topaz-proxy.demo.aserto.com",
+      "OpenFGA": "https://authzen-interop.openfga.dev/stores/01JG9JGS4W0950VN17G8NNAH3C"
     }
   },
   "gateways": {
-    "--Pass Through--": "https://authzen-todo-backend.demo.aserto.com",
+    "--Pass Through--": "http://localhost:8080",
```
ogazitt, line 50: Remnant from testing - should go back to the original URL

interop/authzen-todo-backend/src/pdps.json
```diff
@@ -55,10 +58,7 @@
     "Aserto": "http://authzen-gateway-proxy.demo.aserto.com",
     "Axiomatics": "https://pdp.alfa.guide",
     "Cerbos": "https://authzen-proxy-demo.cerbos.dev",
-    "HexaOPA": "https://interop.authzen.hexaorchestration.org",
```
ogazitt, line 58: You'll need to merge main branch back into yours

interop/authzen-todo-backend/src/server.ts
```diff
@@ -31,10 +33,6 @@ export class Server {
     res.json(config);
   }
 
-  async listGatewayPdps(_: Request, res: Response) {
```
ogazitt, line 34: I believe this is also due to this branch not being synced with main

interop/authzen-todo-backend/src/statefulAuthorizationService.ts
```diff
@@ -0,0 +1,94 @@
...
+  insert(_todoId: string, _userId: string): Promise<void> {
+    return Promise.resolve();
+  }
+
+  delete(_todoId: string, _userId: string): Promise<void> {
+    return Promise.resolve();
+  }
+}
+
+class OpenFGAStatefulAuthorizationService implements StatefulAuthorizationService {
```
ogazitt, line 26: I like this interface-based approach - seems nice and clean. 

It may feel a bit like overkill right now since there's only one stateful implementation, but seems like we want a subdirectory with stateful providers, and have each implementation in its own source file?  `statefulAuthorizationService.ts` can import each of them and the constructor can return the imported implementation.

ogazitt (changes requested):

interop/authzen-todo-backend/src/statefulAuthorizationService.ts
aaguiarz: You mean keeping

`export function createStatefulAuthorizationService( pdpBaseName:string, pdpAuthHeader:string, pdpHeader:string): StatefulAuthorizationService {
  console.log(pdpHeader);
  switch (pdpHeader.toLowerCase()) {
    case "openfga":
      return new OpenFGAStatefulAuthorizationService(pdpBaseName, pdpAuthHeader);
    default:
      return new NullStatefulAuthorizationService();
  }
}`

but having the 'OpenFGAStatefulAuthorizationService' class in a 'stateful/openfga.ts' file?

ogazitt: yes - exactly

aaguiarz: @ogazitt done

ogazitt (approved):

PR #241, opened 2025-03-09
Updates search endpoint paths. (from meeting 25 Feb 2025)

dphhyland:
Changes per meeting notes for Feb 25, 2025

Context should be handled similarly across the different endpoints

The readability of search functions could be improved - perhaps add underscore, etc.
Roland suggests
/access/v1/search/action
/access/v1/search/resource
/access/v1/search/subject
Members on the call approved and it will be added to the spec


ogazitt (changes requested):
@dphhyland can you please remove all the package-lock.json files from all the subdirectories? they all have a yarn.lock file already and we should not have both checked in.

PR #276, opened 2025-04-09
PDP Metdata Discovery feature

identitymonk:
As discuss during the WG meeting on April 8th, here is a first proposal for the PDP Metdata Discovery feature for comments.

It is based on [RFC8414](https://datatracker.ietf.org/doc/html/rfc8414) and [I.D-ietf-oauth-resource-metadata](https://datatracker.ietf.org/doc/draft-ietf-oauth-resource-metadata/).

### DIFF
--- a/api/authorization-api-1_0.md
+++ b/api/authorization-api-1_0.md
@@ -50,6 +50,9 @@ contributor: # Same structure as author list, but goes into contributors
 - name: David Hyland
   org: ID Partners
   email: dave@idpartners.com.au
+- name: Jeff Lombardo
+  org: AWS
+  email: jeffsec@amazon.com
 
 normative:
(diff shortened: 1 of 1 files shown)


identitymonk:
Strange that it is failing on:
```
api/authorization-api-1_0.xml(0): Error: IDREF attribute target references an unknown ID "IANA_well_known_uris", at None
```
while my local kramdown-rfc2629 resolves it.

identitymonk:
Solved `Error: IDREF attribute target references an unknown ID "IANA_well_known_uris"`

local make need value as `IANA_well_known_uris` while pipeline needs it as `IANA.well_known_uris`

davidjbrossard:
Looks good to me

davidjbrossard (approved):

PR #285, opened 2025-04-15
Setup search demo app

alexolivier:
Sets up a react-router 7 app for the search API demos in order to get around the CORS issue if we did it all client side.


identitymonk:
Thanks for the submission. We might need more details on how the selection of the routing is done and how to reference the Search API endpoint(s?) [ see #286 ]. Maybe the usage of the Metadata Endpoint will be useful for that. If so we night need a fourth navigation pane to show the metadata document

interop/authzen-search-demo/app/data/pdps.server.ts
```diff
@@ -0,0 +1,13 @@
+const pdps = [
+  {
+    name: "Cerbos",
+    host: "https://authzen-proxy-demo.cerbos.dev",
+  },
+  {
+    name: "Axiomatics",
+    host: "https://...",
+    auth: "Bearer somevalue",
```
ogazitt, line 9: How do we want to handle this?  backend secrets that are injected in as environment variables?

interop/authzen-search-demo/app/data/subjects.server.ts
```diff
@@ -0,0 +1,15 @@
+const subjects = [
```
ogazitt, line 1: We want another file that is the resource file, right?

interop/authzen-search-demo/app/routes/resource.tsx
```diff
@@ -0,0 +1,292 @@
...
+      id: subject,
+    },
+    action: {
+      name: action,
+    },
+    resource: {
+      type: resource,
+    },
+    context: {},
+    page: {
```
ogazitt, line 86: I think we should leave the paging out of this interop demo (as we agreed in the previous meeting)

interop/authzen-search-demo/app/routes/resource.tsx
```diff
@@ -0,0 +1,292 @@
...
+  try {
+    // make AuthZEN API call to get resources for the subject
+
+    // const headers = new Headers();
+    // headers.append("Content-Type", "application/json");
+    // if (pdp.auth) {
+    //   headers.append("Authorization", pdp.auth);
+    // }
+
+    // const authZENResponse = await fetch(`${pdp.host}/access/v1/resource`, {
```
ogazitt, line 100: I think this should be `.../access/v1/search/resource`

interop/authzen-search-demo/app/routes/resource.tsx
```diff
@@ -0,0 +1,292 @@
...
+    },
+  };
+
+  try {
+    // make AuthZEN API call to get resources for the subject
+
+    // const headers = new Headers();
+    // headers.append("Content-Type", "application/json");
+    // if (pdp.auth) {
+    //   headers.append("Authorization", pdp.auth);
```
ogazitt, line 97: I think we should have a "headers" collection that goes for each PDP, which can be used to route to the correct instance / rule. 

E.g. with Topaz we will try to implement the scenario using the "real system" as opposed to a proxy, and we may need a rule to do tail-evaluation once the ReBAC search returns a candidate set.

interop/authzen-search-demo/package-lock.json
ogazitt, line 1: Do we want npm or yarn here? Everything else in the repo uses `yarn`

interop/authzen-search-demo/app/data/pdps.server.ts
alexolivier: Yup once we have some PDPs that we can test against, then this will read from `PROCESS.ENV....`

interop/authzen-search-demo/app/data/subjects.server.ts
alexolivier: Yup so we can show the sample data once it is defined

interop/authzen-search-demo/package-lock.json
alexolivier: I can switch it but yarn is bit outdated these days

interop/authzen-search-demo/app/routes/resource.tsx
alexolivier: Sure it can go into the pdp config file and read from there

davidjbrossard (approved):

Issue #249, opened 2025-03-10
Example of 6.2.4 needs rework


identitymonk:
While non normative, the example of 6.24 - https://openid.github.io/authzen/#name-sample-response-with-additi would be better as `"decision": "false"` as the `reason` displayed are a `deny`.

baboulebou:
Agreed, the response should be `false` .

ogazitt:
agreed. will incorporate into Draft 04.

PR #287, opened 2025-04-15
draft 04 initial commit

ogazitt:
This commit:
* increments the draft number to 04
* fixes example 6.2.4 to address https://github.com/openid/authzen/issues/249


davidjbrossard (approved):
Changes look good. The example is updated to be in line with the context in the response and the draft number is bumped up.

PR #288, opened 2025-04-15
Revert "PDP Metdata Discovery feature"

ogazitt:
Reverts openid/authzen#276

Upon further review, this PR should have been made with a new version number (04) and also needs to be reviewed at the WG before we merge it.


davidjbrossard (approved):
Reverting changes so we can discuss them in WG before merging

Issue #277, opened 2025-04-12
Inconsistent search routes in 12.1.7


krotscheck:
https://github.com/openid/authzen/blob/2ff35ac9a0dc22359b444c84eeb66c4aa6d6f29c/api/authorization-api-1_0.md?plain=1#L1466

The example in 12.1.7 is inconsistent with the other paths:
```
/search/subject
/resource/search
/search/action
```
Should it be `/search/resource` instead?

ogazitt:
good catch!  Addressed in https://github.com/openid/authzen/pull/289

PR #289, opened 2025-04-15
fix resource search URL inconsistency

ogazitt:
This PR addresses issue #277.


davidjbrossard (approved):
Change is consistent. URL pattern is fine.

Issue #42, opened 2023-07-11
Replace robotic with machine


PieterKas:
I have seen machine identities used to mean all non-human entities (including devices and workloads). Still not perfect, but perhaps better than robotic?

ogazitt:
I believe this is a good improvement in the language we use. https://github.com/openid/authzen/pull/296

PR #296, opened 2025-04-20
changed "robotic" to "machine" to describe machine identities

ogazitt:
This addresses https://github.com/openid/authzen/issues/42 (which I think is reasonable feedback). The term "machine identity" is more prevalent than "robotic identity".


ggebel (approved):
Reviewed change of "robotic" to "machine"

PR #297, opened 2025-04-22
Create a dedicated chapter/section for search and nest all search def…

davidjbrossard:
…initions inside.

As agreed in a previous call, I nested the search API under its own section.


ogazitt:
@davidjbrossard given that we've agreed to defer PE from v1 final, do we still want to do this?  Given that `evaluation` and `evaluations` are both top-level sections, and we only have three more (subject, resource, and action search), perhaps it is more symmetric to have the three search sections still stay as top-level sections?

davidjbrossard:
Yes I would still nest all of search under one roof. It's one overall mechanism.

PR #317, opened 2025-05-13
Search test runner

ogazitt:
This PR adapts the existing test runner, with all of its formatting and features, to the search payload.

It provides more concise and also more accurate test results than the one that is currently in this branch in `./test`.


davidjbrossard:
I am not sure why we have this. I wrote a runner last Friday that works through the tests.

ogazitt:
> I am not sure why we have this. I wrote a runner last Friday that works through the tests.

For three reasons:
1. The AI-generated runner has bugs in it. For example, it assumes a `{ type, id }` structure to the array results which doesn't work with action search results, leading to errors like `Mismatch found: Cannot read properties of undefined (reading 'localeCompare')`.  There are other issues too, including using  deepStrictEqual which is an assert and generates exceptions, leading to inefficient processing.
2. The AI-generated runner emits very verbose output that is difficult to visually inspect for successful vs error cases. 
3. The AI-generated runner does not emit the markdown/html that can be used to generate the report that is put into the website.

I was hoping to help with 2 and 3, but realized that rather than building these for the new runner, it would be trivial to adapt the previous runner to this new scenario.

Issue #278, opened 2025-04-12
Inconsistent use of `reason`, `reason_admin`, and `reason_user`


krotscheck:
https://github.com/openid/authzen/blob/2ff35ac9a0dc22359b444c84eeb66c4aa6d6f29c/api/authorization-api-1_0.md?plain=1#L712

Evaluation response example in deny-on-first-deny doesn't match description above- should it be `reason_admin`, `reason_user`, or should there be an additional `reason` field in section 6.2.4? Or is the `reason` field intended to be a user-defined one as per 6.2.2?

ogazitt:
We need an overhaul of the "reason" section.

vatsalgupta:
+1. I also think that the reason object should be out of context altogether. Context is supposed to provide more details to PEP like for MFA while reason is meant for reasons of denial.

PR #348, opened 2025-07-23
Issue #278 - proposed new wording for context response + examples, in…

baboulebou:
…cluding reasons.

### DIFF
--- a/api/authorization-api-1_0.md
+++ b/api/authorization-api-1_0.md
@@ -407,7 +407,7 @@ In the following non-normative example, the PDP requests a step-up authenticatio
 ~~~
 {: #response-with-step-up-example title="Non-normative Example Response with a step-up request Context"}
 
-Even though this specification doesn't enforce any semantics for the `context` response object, if provided at all, the PEP MUST be able to understand this object. 
+If the PEP does not understand information in the `context` response object in the event of a `decision: true`, the PEP MAY choose to reject the decision.
 
 # Access Evaluations API {#access-evaluations-api}
 


api/authorization-api-1_0.md
```diff
@@ -328,77 +328,74 @@ The following is a non-normative example of a simple Decision:
...
-    }
+    "acr_values": "urn:com:example:loa:3",
+    "amr_values": "mfa hwk"
   }
 }
 ~~~
-{: #response-with-context-example title="Example Response with Context"}
+{: #response-with-step-up-example title="Non-normative Example Response with a step-up request Context"}
+
+Even though this specification doesn't enforce any semantics for the `context` response object, if provided at all, the PEP MUST be able to understand this object.
```
mtrimpe, line 398: This is quite a strong requirement. It should be included at top-level in the description. I do like the idea as it would prevent a lot of negotiation on whether PDP's support certain functionality.

It would make it a lot harder to add fields to the context though as all PEP's must then  be able to understand it.

Either way we should rephrase it from "MUST understand" to what the PEP should do if it *doesn't* understand. Something like this then?

```suggestion
If the PEP does not understand information in the `context` response object the PEP SHOULD interpret the request as having a `"decision"` field with the value`false`. 
```

davidjbrossard: @mtrimpe I disagree here - we cannot force a PEP to fail safe (though it sounds like the right behavior from a security mindset's PoV). Some implementations will want to fail open. The only option is to force the PEP to understand the object or fail using an error code at the discretion of the PEP. @baboulebou's wording reflects this better.

api/authorization-api-1_0.md
```diff
@@ -340,77 +340,74 @@ The following is a non-normative example of a simple Decision:
...
```
davidjbrossard, line 410:
```suggestion
If the PEP does not understand information in the `context` response object in the event of a `decision: true`, the PEP MAY choose to reject the decision.
```

davidjbrossard (approved):
reviewed on the call on August 14th

Issue #343, opened 2025-07-01
Abstract, Introduction, Model and Features sections need to be updated


mtrimpe:
- Abstract section still mentions 'may introduce search'
- Features section doesn't list search features
- NIST reference in Introduction section is not an actual reference 
- Model paragraph talks about "Authorization for the Authorization API" but seems to describe authentication (basic/bearer tokens)

mtrimpe:
Raised PR #350 for this

davidjbrossard:
Fixed in https://github.com/openid/authzen/pull/350

Issue #301, opened 2025-04-24
NIST's ABAC SP 800-162 should be a BibEx reference


identitymonk:
NIST's ABAC SP 800-162 mention in the spec is not an hyperlink to the document.

identitymonk:
```
SP.800-162:
    title: Guide to Attribute Based Access Control (ABAC) Definition and Considerations
    target: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-162.pdf
    author:
    - name: Vincent Hu
      role: editor
      org: NIST
    - name: David Ferraiolo
      role: editor
      org: NIST
    - name: Richard Kuhn
      role: editor
      org: NIST
    - name: Adam Schnitzer
      role: editor
      org: BAH
    - name: Kenneth Sandlin
      role: editor
      org: MITRE
    - name: Robert Miller
      role: editor
      org: MITRE
    - name: Karen Scarfone
      role: editor
      org: Scarfone Cybersecurity
    date: 2014
```

davidjbrossard:
@identitymonk will create a PR for this issue and tie them together.

PR #350, opened 2025-07-29
Updated introductory sections

mtrimpe:
Updated introductory sections to address concerns raised in issues #343 and #301.


davidjbrossard (approved):
Per the meeting on 7/29 we are approving this change.

Issue #337, opened 2025-06-23
Clarify Top-Level 4-Tuple Omission in Bulk Access Evaluation Requests (Section 7.1)


vatsalgupta:
The following sentence in Section 7.1, paragraph 4, about bulk Access Evaluation API Requests is confusing:

"If the evaluations array IS present and contains one or more objects, the top-level subject, action, resource, and context keys (4-tuple) in the request object MAY be omitted."

Readers may misinterpret this to mean that the 4-tuple keys (subject, action, resource, context) within each evaluations object can be omitted, rather than the root-level 4-tuple keys outside the evaluations array. Only by reading Section 7.1.1 does it become clear that root-level keys act as defaults for evaluations entries.

Proposal:
Replace with: “If the evaluations array is present, the root-level subject, action, resource, and context keys (4-tuple) MAY be omitted. If included, these keys provide default values for the corresponding fields in each evaluations entry (see Section 7.1.1).”

Rationale: “Root-level” is clearer than “top-level,” and mentioning defaults inline avoids reliance on Section 7.1.1.

mtrimpe:
In the full text defaults are also mentioned already:

> "If the `evaluations` array IS present and contains one or more objects, the top-level `subject`, `action`, `resource`, and `context` keys (4-tuple) in the request object MAY be omitted. However, if one or more of these values is present, they provide default values for their respective fields in the evaluation requests."

I also think top-level and root-level aren't really meaningfully different.

If we want to clarify this further one thing to add might be to explicitly describe that the combination of top-level keys and evaluations-level keys must be the full 4-tuple somehow.

See PR #351 for a proposal for that

vatsalgupta:
Makes sense!

davidjbrossard:
Addressed in https://github.com/openid/authzen/pull/351

PR #351, opened 2025-07-29
Extended evaluations defaults with completeness requirements

mtrimpe:
Based on the feedback from issue #337 I've extended the explanation of defaults for the Evaluations API with requirements to have a 'complete' 4-tuple for every request. 

This should have been defined already anyhow and hopefully this provides sufficient clarity to address the concerns raised in that issue.


davidjbrossard:
Per the last meeting, we are approving this change (@ggebel and @davidjbrossard )

davidjbrossard (approved):

Issue #352, opened 2025-07-29
HTTPS Binding paths are non-normative


mtrimpe:
For our AuthZEN based logging standard we need an identifier to refer to a *type* of AuthZEN request. At the moment the [HTTPS bindings](https://openid.github.io/authzen/#name-https-binding) seem like the only possible candidate for this.

The paths used in the HTTPS bindings are only included as non-normative example; though.

I'm assuming that we don't expect implementations to have different paths/endpoints; so can we make these paths normative?

I'm not sure if the whole path should be normative or only the postfix (`access/v1/evaluation`, `v1/evaluation` or `/evaluation`)

davidjbrossard:
I'm not sure there's a point in having a normative endpoint if we have the metadata endpoint (which does need to be standardized). Discovery will tell you where the other endpoints sit.

That said, if we do want to make the URLs normative then I vote for `/evaluation` only.

mtrimpe:
Good point. I didn't realize the metadata has URLs per endpoint instead of feature flags/booleans relative to the base PDP URL. 

That might actually be good to mention in the transport/binding examples then. 

And I'll raise another issue for my underlying concern (having some sort of identifier for our AuthZEN based logging standard.)

Issue #358, opened 2025-08-21
Change the metadata endpoint from optional to mandatory in the spec


davidjbrossard:
Section 11 says a PDP may expose a metadata endpoint. Because we decided to not make URIs normative, we also decided to make the metadata endpoint mandatory.

We need to create a PR to update the text in https://openid.github.io/authzen/#name-policy-decision-point-metad to include a MUST. The URI to the metadata endpoint in https://openid.github.io/authzen/#name-obtaining-policy-decision-p must also be made normative.

/.well-known/authzen-configuration

George pointed out that we may want to allow for a tenant identifier to be inserted in the path and we need to check how IETF OAuth does it to be conformant/consistent.

/.well-known/<tenant_id>/authzen-configuration
/<tenant_id>/.well-known/authzen-configuration

See https://datatracker.ietf.org/doc/html/rfc8615

gffletch:
The rules are defined in R[FC 8414](https://datatracker.ietf.org/doc/html/rfc8414) Section 3.

> Authorization servers supporting metadata MUST make a JSON document
   containing metadata as specified in [Section 2](https://datatracker.ietf.org/doc/html/rfc8414#section-2) available at a path
   formed by inserting a well-known URI string into the authorization
   server's issuer identifier between the host component and the path
   component, if any.

Section 3.1

>   If the issuer identifier value contains a path component, any
   terminating "/" MUST be removed before inserting "/.well-known/" and
   the well-known URI suffix between the host component and the path
   component.  The client would make the following request when the
   issuer identifier is "https://example.com/issuer1" and the well-known
   URI suffix is "oauth-authorization-server" to obtain the metadata,
   since the issuer identifier contains a path component:

     GET /.well-known/oauth-authorization-server/issuer1 HTTP/1.1
     Host: example.com

So the IETF correct form is 
<host>/.well-known/authzen-configuration/<tenant_id>

gffletch:
In the context of OAuth/OIDC the metadata documents are discovered based on the `iss` (issuer) of the token. This provides the client with the starting URI from which to find the metadata document. In the case of AuthZen the PEP needs to know the PDP endpoint in order to contact it. How the PEP determines the PDP URI is out of scope for this specification.

That said, once the PEP has the PDP URI, then how it constructs the URI to the metadata document is defined by RFC 8414 and  I believe we should follow that pattern. This pattern allows for multi-tenant PDPs in a consistent way.

[...]

PR #360, opened 2025-08-27
Defined relation between paths in HTTP transport binding and PDP metadata + small fixes

mtrimpe:
This PR adds several changes to prevent the confusion that lead to issue https://github.com/openid/authzen/issues/352 being raised. 

It: 
- Defines that the endpoints in the HTTP transport binding should correspond to those defined in the PDP metadata.
- Fixes capitalization of Content-Type and added it in the example.
- Adds the action endpoint to the IANA registry for PDP metadata (incl. small typo)


davidjbrossard (approved):
We reviewed the changes on the 8/28 call and approved them collectively. Michiel will create another PR to add the recommended URLs as per https://github.com/openid/authzen/issues/358

Issue #325, opened 2025-05-20
Leave more leeway for pagination methods


baboulebou:
Currently the spec assumes a Pagination token for all Search results; i.e.,
```
"page": {
    "next_token": "alsehrq3495u8"
  }

```

But some implementations may support other means, such as "offset + limit" for example. Can we possibly make the pagination token explicitly non-normative, and mention that pagination details are implementation-specific ?

davidjbrossard:
The pagination mechanism in the spec today is applied to:
- Subject search
- Action search
- Resource search

@identitymonk also points out that the context object in the access evaluation API and the response in the access evaluations (boxcar) API could benefit from a pagination mechanism.

We need to factor out the pagination mechanism from each section (currently mentioned 3x in each search API) and perhaps even delegate the specifics of pagination to the underlying implementation. As a result the `next_token` example becomes a MAY rather than a MUST.

Would that impact interoperability? If different implementations have different means of pagination, then how can interoperability be achieved? Perhaps the `page` field is part of the spec and the contents is either a `next_token` or the combination of `offset` and `limit` as suggested by @baboulebou .

Moving to the 6/24 meeting for debate.

davidjbrossard:
- Consistency vs. snapshot isolation

If we return metadata about the search, the metadata may be about the search at the original time of execution and that could be updated. We're not requiring implementations to achieve snapshot isolation to guarantee correctness or return the full set.

davidjbrossard:
Note: use SCIM as inspiration

mtrimpe:
Raised PR with proposal for typed navigation: #374

davidjbrossard:
On a call today, the chairs have decided to retain cursor-based pagination only in the core spec. We leave open the option to create a later profile that may offer alternative pagination methods such as offset-based. The consensus is that clients will be easier to write and maintain and interoperate with PDPs if we have a single consistent way of handling pagination.

PR #374, opened 2025-09-15
Adding first draft of typed pagination

mtrimpe:
Added the first draft of the typed pagination approach to address #325.

When reviewing pay extra attention to the "Paginated Requests" section as that's where the most design decisions needed to be made.

I also adjusted and cleaned up some of the IANA Considerations after diving into the expectations there.

### DIFF
--- a/api/authorization-api-1_0.md
+++ b/api/authorization-api-1_0.md
@@ -1772,11 +1772,9 @@ The reason for the use of the mailing list is to enable public review of registr
 
 ## AuthZEN Policy Decision Point Metadata Registry {#iana-pdp-metadata-registry}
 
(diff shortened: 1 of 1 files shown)


api/authorization-api-1_0.md
```diff
@@ -1059,17 +1025,21 @@ The following payload defines a request for the resources of type `account` on w
...
-## The Resource Search API Response {#resource-search-response}
-
-The response is a paged array of Resources.
+{: #search-pagination-token-initial-request title="Example initial Search Request with token-based pagination"}
 
 ~~~ json
 {
+  "page": {
+    "next_token": "a3M9NDU2O3N6PTI=",
+    "count": 2
```
gertd, line 1041: Do we need `count` if the length of the `results` array provides the real number?

PR #378, opened 2025-09-24
Token-based pagination with support for capabilities

mtrimpe:
In this PR I'm offering up another proposal which adopts `token` based pagination as the default; but defines `limit` and `offset` capabilities which, when combined, can also support classic page-based navigation.


api/authorization-api-1_0.md
```diff
@@ -888,165 +889,97 @@ The following is a non-normative example of a response to an Access Evaluations
...
-: REQUIRED. The subject (or principal) of type Subject.  NOTE that the Subject type is REQUIRED but the Subject ID can be omitted, and if present, is IGNORED.
+Search APIs can return large result sets. To manage this, a Policy Decision Point (PDP) MAY support pagination, allowing clients to navigate and retrieve subsets of the total result set. 
 
-`action`:
-: REQUIRED. The action (or verb) of type Action.
+Pagination does not guarantee an atomic snapshot of the result set. Consequently, if items are added or removed while paginating, results MAY be repeated or omitted between pages.
 
-`resource`:
-: REQUIRED. The resource of type Resource.
+Pagination is based on the use of opaque tokens. A client makes an initial request for data by sending a query that does not contain a token. In response, the PDP MAY initiate pagination by returning a partial result set instead of the complete one.
```
ogazitt, line 911: Pagination is based on the use of opaque tokens. A client makes an initial request for data by sending a query that does not contain a token. If the PDP determines that the response contains too many results to fit in a single response, the PDP returns a partial result set and a token that the caller can use to retrieve the next page of results.

api/authorization-api-1_0.md
```diff
@@ -888,165 +889,97 @@ The following is a non-normative example of a response to an Access Evaluations
...
-`context`:
-: OPTIONAL. Contextual data about the request.
+A paginated response MUST be clearly identified by the inclusion of a `page` object containing a non-empty, opaque `next_token`. This token is the signal to the client that more results are available.
 
-`page`:
-: OPTIONAL. A page token for paged requests.
+To retrieve the next page, the client sends a subsequent request containing a `page` object with the `token` field set to the `next_token` value from the previous response. This process is repeated until the PDP returns a `page` object in which the value of the `next_token` field is an empty string, signaling the end of the result set.
 
-### Example (non-normative)
+Additionally, a PDP MAY allow clients to influence pagination with parameters such as `limit` and `offset`. Support for these capabilities MUST be advertised in the PDP's `supported_capabilities` metadata (see {{pdp-metadata-data-capabilities}}).
```
ogazitt, line 917: Clients of the search APIs that wish to sequentially iterate through the entire result set SHOULD use the core pagination mechanism described above, which is designed to work consistently across all PDPs that support the search APIs.

### Advanced Pagination

Some PDPs MAY allow clients to retrieve specific pages of results by passing in parameters such a `limit` and `offset` to instruct the PDP to return a specific slice of the result set. Support for this advanced capability MUST be advertised in the PDP's `supported_capabilities` metadata (see {{pdp-metadata-data-capabilities}}).

api/authorization-api-1_0.md
```diff
@@ -888,165 +889,97 @@ The following is a non-normative example of a response to an Access Evaluations
...
 
-# Resource Search API {#resource-search-api}
+`properties`:
+: OPTIONAL. An object containing additional pagination response attributes. Examples include, but are not limited to, estimated totals or the number of remaining results.
 
-The Resource Search API defines the message exchange pattern between a client (PEP) and an authorization service (PDP) for returning all of the resources that match the search criteria.
+### Offset capability {#search-pagination-offset}
 
-The Resource Search API is based on the Access Evaluation information model, but omits the Resource ID.
+Using the offset capability, clients can indicate the number of results to skip before starting to return results. This enables clients to interactively navigate result sets and, when combined with the Limit capability ({{search-pagination-limit}}), can support classic page-based navigation.
```
ogazitt, line 953: My general comment is that the spec should cleanly separate the base mechanism (paging using opaque tokens, which is guaranteed to work across implementations), and the "advanced paging" mechanism.  The former should have non-normative examples. The latter should have non-normative examples in a separate section.  

This makes it much easier for spec readers to implement a client of their choosing.

ogazitt (changes requested):

api/authorization-api-1_0.md
mtrimpe: Resolved by omitting offset

PR #379, opened 2025-10-01
Added endpoint overview and JSON mapping

mtrimpe:
Changed HTTPS binding to use a generic endpoint overview instead of repeating instructions per endpoint.
Made Information Model transport/JSON agnostic
Introduced JSON mappings in the HTTPS transport binding.


davidjbrossard (approved):
Thanks for the massive reorg and cleanup, Michiel.

PR #387, opened 2025-10-16
Change IdP demo to record

alexolivier:
Migrate the demo app to use the search interop demo https://hackmd.io/@oidf-wg-authzen/idp-integration
Screenshot 2025-10-16 at 21 53 25


ogazitt (commented):
Looks good!

The only thing I find a bit unfortunate is that the code uses 4 spaces per tabstop, whereas the rest of the repo samples use 2 spaces per tabstop (for javascript/typescript/JSON). This ends up manifesting most obviously in the JSON, which this PR reformatted. 

It's not that big of a deal, so I'm approving, but ideally we'd have consistent lint rules for JS/TS?

ogazitt (approved):

PR #394, opened 2025-10-24
Add Curity IDP

iggbom:
- Added option to select Curity as an IDP. The authorize and token endpoints are resolved from the Metadata URL that is in turn resolved from defining the issuer.
- Added drop-down to select IDP to use for Login.

Needs the following ENV:
- ISSUER
- BASE_URL
- CLIENT_ID
- CLIENT_SECRET
- SCOPE (optional)
- PDP_CONFIG


alexolivier:
Thanks for the push to use discovery, I've incorporated this into https://github.com/openid/authzen/pull/395 - closing this PR.

PR #449, opened 2026-03-11
Feature/obligations

baboulebou:
Proposal for obligations support in Authzen response context...


alexolivier:
@baboulebou what should we do with this?

davidjbrossard:
@baboulebou can you resolve the conflicts with the core spec?

davidjbrossard:
Should this PR either be changed to a separate profile or rejected altogether? This made changes and proposed a 1.1 but this isn't what we decided lately (have obligations as a profile). Am I mistaken?

baboulebou:
Looks like this was an old one. Closing it.

PR #541, opened 2026-06-25
COAZ: split into framework + MCP profile, rework mapping model

### LINKED ISSUE #485: Is coaz:true required?
From @maxwellgerber:

If x-coaz-mapping is required, do you still need the coaz: true field on the response?

### LINKED ISSUE #491: CEL static string syntax is error-prone and inconsistent with common schema conventions
The spec requires every string value in x-coaz-mapping to be a CEL expression, with static strings wrapped in single quotes:
                                                                                                                                                                                          
  "type": "'customer'"         ← static literal
  "id": "params.arguments.id"  ← dynamic expression                                                                                                                                       
                                                                                                                                                                                          
  This is a footgun. If a developer forgets single quotes on a static value:                                                                                                              
                                                                                                                                                                                          
  "type": "customer"  ← CEL interprets this as a variable named "customer", not the string "customer"                                                                                     
                                                                                                                                                                                          
  This fails silently or with a confusing CEL evaluation error, with no indication at the schema level that anything is wrong. No widely used schema format works this way — JSON Schema, 
  OpenAPI extensions, and similar standards distinguish static values from expressions structurally, not syntactically.                                                                   
                                                                                                                                                                                          
  **Suggested Resolution**                                                                                                                                                                    
   
  Use a structural distinction instead:                                                                                                                                                   
                  
  "type": { "value": "customer" }          ← static                                                                                                                                       
  "id":   { "expr": "params.arguments.id" } ← dynamic                                                                                                                                     
                                                                                                                                                                                          
  This makes intent explicit, is easier to validate at parse time, and is consistent with how other schema extension mechanisms work.

### LINKED ISSUE #493: Explain why the COAZ mapping is needed
From @ogazitt in the AuthZen call on 4/16.
Most MCP developers may not be familiar with the AuthZen model, so it makes sense for the spec to explain why it is needed.

### LINKED ISSUE #494: Use evaluation/evaluations syntax instead of arrays
From @ogazitt in the AuthZen call on 4/16:

In specifying the COAZ mapping, make the mapping look exactly like how we pass parameters to evaluation/evaluations call. That way we can avoid specifying arrays and then converting that to what an evaluation / evaluations call needs.

alexolivier:
## Summary

**These are the proposed changes coming out of the OpenID AuthZEN WG call on 2026-06-25, updated for the review feedback so far.**

The single COAZ-for-MCP document is split into two:

1. **`authzen-coaz-framework-1_0.md` — the COAZ framework (new).** A protocol-neutral framework: COAZ is the mapping of *any* information model (HTTP request, MCP JSON-RPC message, OpenAPI route, …) into an AuthZEN Authorization API request. It defines the request envelope, the value model (literals vs expressions), the expression contract with CEL as the default language, trust-anchored fields, optional default/declared mappings, and a **binding conformance checklist** that concrete bindings fill in. It defines no protocol-specific mapping itself.

2. **`authzen-mcp-profile-1_0.md` — COAZ-MCP (reworked).** The COAZ binding for MCP.

## Key changes

- **Framework / binding split** so HTTP and OpenAPI bindings can sit alongside COAZ-MCP later — addresses the "explain why the mapping is needed" ask by giving COAZ a generic conceptual home (**#493**).
- **Request envelope** replaces the SARC-arrays + cardinality machinery: a mapping is a single outer key naming the AuthZEN API to call. `evaluation` wraps a single-decision Access Evaluation request; every default mapping uses it, so the bar for a PDP is exactly AuthZEN's own minimum. `evaluations` wraps a multi-decision Access Evaluations request, engaged only by declared mappings that need several checks for one call (a PEP whose PDP lacks the batch API may fan the entries out as individual evaluation calls). Unknown envelope keys fail closed, leaving room for future keys such as resource search (**#494**).
- **New value syntax**: bare values are literals, `$`-prefixed strings are CEL expressions, `$$` escapes a literal `$` — replacing the error-prone single-quoted-CEL-literal convention (**#491**). The framework names CEL as its default expression language (SHOULD); a binding may designate another with justification.
- **`coaz: true` marker removed**: the presence of `x-authzen-mapping` is the sole signal a tool declares a mapping (**#485**).
- **Scope broadened to all MCP methods** (not just `tools/call`) via a per-method **default mapping table**, with an explicit pass-through set (`ping`, `notifications/*`) and a fail-closed rule for unknown methods.
- **Gateway or server as PEP**: the architecture covers both deployment shapes. The mapping travels in the `tools/list` response, so a gateway can enforce server-authored mappings with no out-of-band configuration, and a server without a gateway enforces its own.
- **Subject identity established by the PEP from the token** — `sub` by convention, with an optional deployment-designated on-behalf-of claim — and **verified** by the PEP so a server-authored declared mapping cannot assert a different principal. Declared mappings may shape `subject.type`/attributes, `action`, `resource`, and `context`, all treated as untrusted input.
- **Denials and errors distinguished**: a deny decision is the normal course of policy enforcement, not a fault; mapping errors and PDP communication failures are the error categories. All three fail closed.

## Closes

- Closes #485 — Is `coaz:true` required?
- Closes #493 — Explain why the COAZ mapping is needed
- Closes #494 — Use evaluation/evaluations syntax instead of arrays
- Closes #491 — CEL static string syntax is error-prone

## Open questions for the WG

- **Naming — settled in review**: concrete instantiations are *bindings* designated COAZ-\<protocol\> (this document is now "COAZ-MCP: COAZ Binding for the Model Context Protocol"), and the extension field name stays `x-authzen-mapping`. Still open: whether file names / docnames / published URLs should follow the COAZ-MCP designation (currently unchanged to preserve links).
- **Notifications**: currently all `notifications/*` are pass-through (no `id`, so no JSON-RPC error channel). Should there be an interoperable opt-in for deployments that want notifications authorized?
- **Subject trust anchor vs agents that can't carry the user in the token** (e.g. a customer-service bot acting on an emailed request): current draft anchors `subject.id` to the token and models the end customer as context; an alternative is "at least one subject field from the token". To discuss on a WG call.
- **Per-tool authorization opt-out** (`coaz:not_required`): whether a server can mark a tool as exempt, or whether bypass should stay a policy/PEP-config decision.
- **Draft numbering** — whether splitting into a new document should bump the draft number, and how to version the framework and bindings independently.


profiles/authzen-mcp-profile-1_0.md
```diff
@@ -391,637 +303,579 @@ Given the following `tools/call` request:
...
-          }
-        }
-      }
-    ]
-  }
-}
+// tools/call  (applies when the tool declares no mapping)
+{ "subject": { "type": "identity", "id": "$token.sub" },
+  "context": { "agent": "$token.?client_id" },
+  "evaluations": [ { "action": { "name": "$params.name" },
```
ogazitt: should thiis be `{ "action": { "name": "tool/invoke" } ... }` instead of the name being `"$params.name"` so as not to be redundant with `{ "resource": { "type": "tool", "id": "$params.name" } }`?

ogazitt (commented):
A few general comments and a specific question in the commit diff.
1. I think the factoring achieved its intended purpose in the sense that there is a general framwork and specific bindings (MCP being one of them). Reading the language, the general framework is a bit "academic" in the language. But I think it formalizes some useful aspects which is very helpful - specifically the notion of default mappings, declared mappings, and discoverability.
2. I wonder if the framework needs to get into whether the specific bindings need to declare support for a specific expression language. Unless the general framework mandates an expression language (like CEL) and then allows the individual bindings to override it, is it useful to list it as a requirement that each binding needs to provide?
3. Should we utilize the term "binding" to refer to a specfiic instantiation of the COAZ framework (i.e. the COAZ MCP binding vs the COAZ MCP profile)?
4. I don't love the fact that the framework mandates the use of `evaluations`. I think that the default binding will likely always be `evaluation` so why would we set the bar at an optional capability (`evaluations`)?  I think providing choice is easily handled by using an outer key name that specifies the call to make - something like ```{ "evaluation": { "subject": { ...} } }`.  This also provides future extensibility - e.g. if we want to provide the facility to filter a response to `tools/list` that uses the AuthZEN resource-search API...
5. I don't fully understand the "trust anchors" discussion. If I follow the logic, the concern here is that a server may send some kind of bogus mapping that flips the subject to something different than what the caller would expect (JWT.sub claim). But the MCP server could do that to any of the fields... and isn't the security model here more focused on the MCP server being the authoritative source of how it requires its tool invocations and resource requests to be authorized?  Stated otherwise, if the MCP server believes that the correct `subject.id` comes from a different part of the request, doesn't it know best?  The client here is the agent, and it's untrusted from the MCP server's perspective - just like any API client is untrusted from an API server's perspective. What am I missing here?

embesozzi:
I 100% agree that it should be agnostic and support the concept of binding. Nowadays, I use the concept of Agentic Profiles. These could be MCP Profiles, Agent Cards for A2A, or even just an OpenAPI specification, allowing an agent to discover the required permissions (SARC model) based on an idea similar to the COAZ mapping. This also helps implement the concept of intent-based authorization.

alexolivier:
Thanks @ogazitt - point by point:

**1. Academic language.** Fair. The framework doc drifted abstract while I was pulling the MCP specifics out of it. I'll do a pass to ground it with concrete examples so it reads less like a paper.

**2. Expression language.** Agree, cleaner your way. I'll have the framework name CEL as the default and let a binding override it. MCP stays on CEL, so nothing changes in practice, it just stops looking optional when it isn't.

**3. "Binding" vs "profile".** Agree, let's standardise on binding for a concrete instantiation (COAZ MCP binding). I'll sweep the docs. "Profile" kept leaking in from the original single-doc naming.

**4. Mandating `evaluations`.** A single check is just a one-element `evaluations` request, so mandating it doesn't cost the common case anything, you write the same thing with one entry in the array. What it buys is one shape across every method, including the ones that fan out to multiple checks, without the mapping author having to decide which endpoint they're hitting.

The outer-key envelope (`{ "evaluation": {...} }` vs `{ "evaluations": {...} }`) is where I'm not so sure. It hands the mapping author a decision about what kind of call to make, and I'd rather the binding pin that down than leave it open per-tool.

On resource-search for filtering `tools/list`, I don't think the envelope helps. Resource search is a different endpoint entirely, not something you'd select by swapping a verb key in the same body. So `tools/list` filtering is a separate binding rule when we get to it, not an `evaluation` variant.

I think batch should just be table-stakes for anything implementing COAZ, but I'm open to being argued out of it if people think it excludes too many PDPs. 

**5. Trust anchors.** You're mostly right I've likely overcooked this section.

If a gateway is brokering the connection, that gateway has to be trusted anyway, as it's calling the PDP. And if the MCP server is the PEP, it authenticated the caller itself, so it does know best, exactly as you say. Either way the _thing_ doing the mapping is something we're already trusting. The scenario I was defending against (a server flipping `subject.id` to impersonate someone) mostly goes away if we require you to accept the PEP is trusted - but does that break zero trust?

I think its worth keeping a one-liner that subject identity conventionally derives from the validated token (`sub` by default, an on-behalf-of claim where a deployment designates one), because that's a useful convention, but I'll drop the framing that treats the mapping author as hostile.

**Inline (`action.name`):** Good catch, agreed.

Thanks @embesozzi, this is the direction I want to push too. OpenAPI is probably the next one I'd tackle after MCP, highest-leverage for adoption since so much is already described that way. A2A and Agent Cards fit the same shape. Intent-based authz is an interesting layer on top, keen to see how your Agentic Profiles model lines up with the SARC mapping here.

ogazitt:
@alexolivier sounds like we agree on everything except for:

> **4. Mandating `evaluations`.** A single check is just a one-element `evaluations` request, so mandating it doesn't cost the common case anything, you write the same thing with one entry in the array. What it buys is one shape across every method, including the ones that fan out to multiple checks, without the mapping author having to decide which endpoint they're hitting.
> 
> The outer-key envelope (`{ "evaluation": {...} }` vs `{ "evaluations": {...} }`) is where I'm not so sure. It hands the mapping author a decision about what kind of call to make, and I'd rather the binding pin that down than leave it open per-tool.
> 
> On resource-search for filtering `tools/list`, I don't think the envelope helps. Resource search is a different endpoint entirely, not something you'd select by swapping a verb key in the same body. So `tools/list` filtering is a separate binding rule when we get to it, not an `evaluation` variant.
> 
> I think batch should just be table-stakes for anything implementing COAZ, but I'm open to being argued out of it if people think it excludes too many PDPs.
> 

My biggest concern with mandating `evaluations` is that it is an optional capability with AuthZEN 1.0, so we are setting the bar higher than the minimum that AuthZEN requires of PDPs.  This would be reasonable if the common case for the COAZ MCP binding required multiple AuthZEN evaluations. But I believe the common case is actually a single `evaluation`.  There is only one example in the profile where you're using more than one array element, and that's when you have tool that operates over two resources and needs to evaluate a permission (read / write) for both resources.

That is, of course, a valid scenario, and that's why I believe we should support `evaluations` as an option in the COAZ MCP binding. But I'm not sure we should require it? 

And to me, the way to indicate which AuthZEN API you're using is very easy to encode in the COAZ MCP binding - simply an outer key. 

I take your point that filtering is a different scenario than making a binary authorization decision. I do think it's valuable in the `tools/list` scenario, so it's worth capturing as a requirement. But I'm NOT advocating that we gate publishing an updated draft on providing this capability.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+Declared Mapping:
+: A mapping supplied by an authority that describes an operation (for example,
+  a server advertising a mapping for a specific operation). OPTIONAL; see
+  {{default-declared}}.
+
+# The COAZ Model {#model}
+
+## Overview
+
+A COAZ mapping is a template, shaped exactly as an AuthZEN Access Evaluations
```
tulshi: A COAZ mapping is a template, corresponding to a distinct operation. It is shaped exactly as an AuthZEN Access Evaluations request, in which...

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+Information Model:
+: The set of data associated with an operation that a profile makes available
+  to a mapping, such as request fields, message parameters, and security
+  tokens.
+
+Input Variable:
+: A named element of the information model, exposed to expressions by a profile
+  (for example, a variable holding a request's parameters or a variable holding
+  decoded token claims).
+
```
tulshi, line 180: Operation: A unit of work in the specific protocol or interface, which requires an independent authorization decision from the AuthZEN PDP

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+protocol might expose a variable for the request and a variable for the
+caller's authorization token.
+
+A profile MUST specify, for each input variable, its name, the source from
+which the PEP populates it, and its structure, so that mapping authors can
+write expressions against it deterministically.
+
+## The Mapping {#mapping}
+
+A COAZ mapping is a JSON {{RFC8259}} object structured as an AuthZEN Access
```
tulshi: Add a reference to Section 7 (Access Evaluations API) of the AuthZen spec here.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+interoperable way to obtain such decisions using the
+Subject-Action-Resource-Context (SARC) model. What it does not define
+is how to *derive* an AuthZEN request from the information model of any
+particular protocol.
+
+COAZ (Compatible with OpenID AuthZEN) is the framework that fills this gap. It
+defines a single, protocol-neutral pattern:
+
+~~~ ascii-art
+   any information model        a declarative mapping        an AuthZEN request
```
tulshi: Instead of "any information model" -> "An operation in any information model"

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+which the PEP populates it, and its structure, so that mapping authors can
+write expressions against it deterministically.
+
+## The Mapping {#mapping}
+
+A COAZ mapping is a JSON {{RFC8259}} object structured as an AuthZEN Access
+Evaluations request. It MAY contain the top-level `subject`, `action`,
+`resource`, and `context` fields, and an `evaluations` array whose entries MAY
+each contain their own `subject`, `action`, `resource`, and `context` fields.
+
```
tulshi: It'll be great to introduce a few examples at this point in the draft. One with single SARC expressions, one with an array expression. See Section 5 in the SSF spec for reference: https://openid.net/specs/openid-sharedsignals-framework-1_0-final.html#section-5

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+The framework does not redefine this structure or its semantics. The meaning of
+the top-level fields, the meaning of the per-entry fields, and the rules by
+which per-entry fields override top-level defaults are exactly as defined for
+the Access Evaluations request in {{AUTHZEN}}. In particular, custom attributes
```
tulshi: Refer to Section 7, not the whole spec.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+any particular syntax for this discriminator or escape; it requires only that
+the distinction be lossless.
+
+This framework does not use the terms "static" or "dynamic" to describe values.
+A field reference such as "the request's path" is an expression, not a literal,
+because it is computed from an input variable. Only constants are literals.
+
+## The Expression Contract {#expression-contract}
+
+A profile MUST designate an expression language (for example, Common Expression
```
tulshi: I don't mind if we standardize on CEL here. If it isn't appropriate for some specific protocol / interface, we can define another version of this draft that has that as the specialized mapping (e.g. COAZ-JSONPath). IMO, keeping the CEL part abstract here gives too much flexibility and weakens this draft.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+  identify any operations that are *not* authorized by COAZ (pass-through
+  operations), so that the absence of a mapping is never interpreted as an
+  implicit deny.
+
+- **Declared mappings.** A profile MAY allow an authority that describes an
+  operation to supply a mapping for it. A profile that supports declared
+  mappings MUST define who may author them, where they are carried, and how a
+  declared mapping relates to any default mapping for the same operation (for
+  example, whether it overrides the default).
+
```
tulshi, line 452: An example of a declared and default mapping will be useful here.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+defines the categories and their meaning; the transport used to report each
+category to the caller is profile-defined, because it depends on the protocol.
+
+Mapping Error:
+: The PEP cannot construct a valid AuthZEN request from the mapping and the
+  operation's inputs — for example, an expression references a non-existent
+  field, an expression fails to evaluate, the mapping is malformed, or a
+  required field is missing. The PEP MUST NOT permit the operation.
+
+Authorization Denial:
```
tulshi, line 540: Is denial really an error? I think it's a normal course of operation.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,573 @@
...
+: The PDP returns a deny decision for one or more evaluations. The PEP MUST NOT
```
tulshi: So I guess we don't want "fail open" implementations? Those are useful, often when things are still in development.

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -67,53 +78,55 @@ normative:
...
 
 --- abstract
 
-This specification defines a profile of the OpenID AuthZen Authorization API for use with the Model Context Protocol (MCP). It introduces COAZ (Compatible
-with OpenID AuthZen), a standardized mapping from MCP tool definitions and
-their invocation parameters to the AuthZen Subject-Action-Resource-Context
-(SARC) model. This enables MCP gateways and servers to perform
-fine-grained, parameter-level authorization checks via an AuthZen Policy
-Decision Point (PDP) before executing MCP tools. It also helps MCP Clients understand how authorization will be performed by the server, so that it can provide appropriate parameters to the MCP server tools it invokes.
+This specification is a COAZ profile, as defined by the COAZ framework
```
tulshi: framework -> Framework

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -47,6 +47,17 @@ normative:
         name: Atul Tulshibagwale
         org: SGNL
     date: 2026
+  COAZFW:
```
tulshi, line 33: Let's use the acronym "COAZ" for the framework. Each profile can be designated after a dash, like COAZ-MCP.

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -67,53 +78,55 @@ normative:
...
 
-This specification defines a profile of the OpenID AuthZen Authorization API for use with the Model Context Protocol (MCP). It introduces COAZ (Compatible
-with OpenID AuthZen), a standardized mapping from MCP tool definitions and
-their invocation parameters to the AuthZen Subject-Action-Resource-Context
-(SARC) model. This enables MCP gateways and servers to perform
-fine-grained, parameter-level authorization checks via an AuthZen Policy
-Decision Point (PDP) before executing MCP tools. It also helps MCP Clients understand how authorization will be performed by the server, so that it can provide appropriate parameters to the MCP server tools it invokes.
+This specification is a COAZ profile, as defined by the COAZ framework
+{{COAZFW}}, for the Model Context Protocol (MCP) {{MCP}}. It defines how MCP
+JSON-RPC messages are mapped into requests to the OpenID AuthZEN Authorization
```
tulshi: "requests to the OpenID AuthZEN" -> "Access Evaluations API calls of the OpenID AuthZEN"?

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -67,53 +78,55 @@ normative:
...
-interface for dynamic, fine-grained authorization decisions using the
-Subject-Action-Resource-Context (SARC) model. This specification defines a
-profile of AuthZen for MCP tool invocations, enabling MCP participants to
-leverage AuthZen PDPs for access control decisions that go beyond what OAuth
-alone can express.
+and invoke capabilities offered by MCP servers — tools, resources, prompts, and
+more — over JSON-RPC {{JSONRPC}}. Authorization in MCP relies on OAuth 2.1
+{{OAUTH21}}, but OAuth alone leaves some concerns unaddressed:
+
+- The OAuth access token is issued to the agent, which may act autonomously on
```
tulshi: **If** the OAuth access token is issued to the agent...

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -67,53 +78,55 @@ normative:
...
-The OpenID AuthZen Authorization API {{AUTHZEN}} provides a standardized
-interface for dynamic, fine-grained authorization decisions using the
-Subject-Action-Resource-Context (SARC) model. This specification defines a
-profile of AuthZen for MCP tool invocations, enabling MCP participants to
-leverage AuthZen PDPs for access control decisions that go beyond what OAuth
-alone can express.
+and invoke capabilities offered by MCP servers — tools, resources, prompts, and
+more — over JSON-RPC {{JSONRPC}}. Authorization in MCP relies on OAuth 2.1
+{{OAUTH21}}, but OAuth alone leaves some concerns unaddressed:
+
```
tulshi: I think we should flip the order of these two points.  The "scopes" point is more important than "finding the user id if the token is issued to the agent" point.

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -67,53 +78,55 @@ normative:
...
+  scopes cannot express.
+
+The OpenID AuthZEN Authorization API {{AUTHZEN}} provides standardized,
+fine-grained authorization using the Subject-Action-Resource-Context (SARC)
+model. The COAZ framework {{COAZFW}} defines, in a protocol-neutral way, how to
+project the information model of a protocol into an AuthZEN Access Evaluations
+request. This document is the COAZ *profile* for MCP: it binds that framework to
+MCP's information model.
+
+This profile authorizes **all** MCP messages, not only tool invocations. It does
```
tulshi: You can drop the "not only tool invocations" part, since the tool mapping never saw the light of day, so this is all new work.

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -124,265 +137,164 @@ when, and only when, they appear in all capitals, as shown here.
...
-  session tokens map to the AuthZen SARC model.
+| Framework requirement | This profile |
+|:---|:---|
+| Information model | `params`, `token` ({{information-model}}) |
+| Mapping location | `x-authzen-mapping` in a tool's `inputSchema`; otherwise the default mapping ({{declared-mappings}}, {{default-mappings}}) |
+| Literal/expression discriminator | bare value = literal; `$`-prefixed string = CEL expression; `$$` escape ({{expressions}}) |
+| Expression language | Common Expression Language {{CEL}} ({{expressions}}) |
+| Operations in scope | All MCP methods, per the default mapping table; pass-through set listed ({{default-mappings}}) |
+| Default mapping behavior | Per-method default mapping table ({{default-mappings}}) |
+| Declared mapping behavior | MCP server MAY declare a mapping for a tool; overrides that tool's default ({{declared-mappings}}) |
```
tulshi: It might be clearer to say "overrides the default for that tool..." instead of "overrides that tool's default...".

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -124,265 +137,164 @@ when, and only when, they appear in all capitals, as shown here.
...
-specification to construct the AuthZen request parameters from the tool call
-arguments and the associated access token.
-
-It is also possible for the MCP client to call the AuthZen PDP before making
-the tool call. However, since the MCP client may not be trusted by the MCP
-server, this does not preclude the MCP server from performing its own
-authorization check.
+The response to `tools/list` carries each tool's `inputSchema`, and, for tools
+that declare a mapping, the `x-authzen-mapping` within it. Carrying the mapping in
+`tools/list` makes it available to the MCP client, so that the client (or the
```
tulshi: I think making it available to the MCP Gateway (if one exists) is more important than making it available to the client. Without this mapping, the MCP Gateway cannot act as a PEP. Since a Gateway may be managed by a generic IT team, they may not know enough about the business logic of a specific MCP server behind it to define the mapping themselves (at the Gateway). Therefore it is better to define the mapping in the MCP Server, and let the Gateway enforce it.

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -124,265 +137,164 @@ when, and only when, they appear in all capitals, as shown here.
...
+A CEL expression MUST evaluate to a single JSON value — a scalar, list, or map —
+as required by the COAZ expression contract ({{COAZFW}}, Section 3.5). A list or
+map returned by an expression is a single field value and does not, by itself,
+produce multiple evaluations.
+
+CEL field selection on a missing key is an error. To populate an OPTIONAL field
+from a claim or parameter that may be absent, an expression MUST use CEL optional
+selection (the `.?` operator, e.g. `$token.?client_id`); an expression that
+yields an absent optional causes its field to be omitted from the request, as
+defined by the COAZ expression contract. An expression that errors, or that
```
tulshi: I'd like to replace "An expression that errors" with a more formal "An expression whose evaluation results in an error"

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -124,265 +137,164 @@ when, and only when, they appear in all capitals, as shown here.
...
+yields absent or null for a REQUIRED field (`subject`, `action`, or `resource`),
+is a mapping error ({{mapping-errors}}). Because evaluating CEL requires a CEL
```
tulshi: I'm not sure I understand this line ("Because evaluating CEL ...")

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -391,637 +303,579 @@ Given the following `tools/call` request:
...
 
-And an access token with the following decoded claims:
+## Pass-through Operations
+
+The following are pass-through: the PEP MUST NOT call the PDP for them and MUST
+allow them to proceed. They are listed explicitly so that the absence of a
+mapping is never interpreted as a deny.
+
+- `ping`
+- all notifications (`notifications/*`), which carry no `id` and expect no
```
tulshi: fine to not enforce ping, but what if an MCP server requires the client to authorize every notification with an OAuth token? Can we enforce that?

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -391,637 +303,579 @@ Given the following `tools/call` request:
...
+          ]
+        }
       }
     },
-    "context": {
-      "type": "array",
-      "description": "Array of contextual information relevant to the authorization decision",
-      "items": {
+    {
+      "name": "get_local_weather",
```
tulshi: Based on the current content of Section 6.9, the "get_local_weather" will be treated as an error because it lacks a mapping? In my proposal I had a line "coaz:true" for those tools that needed a mapping. I think it should be OK to allow tools that don't require mappings.

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -391,637 +303,579 @@ Given the following `tools/call` request:
...
+The following are pass-through: the PEP MUST NOT call the PDP for them and MUST
+allow them to proceed. They are listed explicitly so that the absence of a
+mapping is never interpreted as a deny.
+
+- `ping`
+- all notifications (`notifications/*`), which carry no `id` and expect no
+  response.
+
+## Unknown Methods
+
```
tulshi: There is merit to ensuring that every callable element (tool, resource, prompt, etc.) should have a mapping. However, I'd like it if some of those elements can be declared as not having a mapping at all. 

In my proposal, only the tools which had "coaz:true" would be evaluated, but in this proposal, perhaps we can define it negatively? So for a specific tool / resource /prompt, one can say "coaz:not_required"?

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -391,637 +303,579 @@ Given the following `tools/call` request:
...
+An MCP server MAY declare a mapping for a tool by including `x-authzen-mapping` in
+the tool's `inputSchema` ({{declaring-support}}). A declared mapping has the same
+shape as a default mapping — an AuthZEN Access Evaluations request template — and
+uses the same expression and literal rules ({{expressions}}).
+
+A declared mapping for a tool **overrides the default `tools/call` mapping for
+that tool only**. It does not affect the default mapping for any other method or
+for `tools/call` of any other tool.
+
+The subject identifier (`subject.id`) is a trust-anchored field ({{COAZFW}},
```
tulshi: I can think of scenarios where the agent is acting on behalf of a user, but has no means of embedding the user id in the token. For example, a customer service bot responding to a customer email. This should be allowed. The agent should be able to pass the subject in one of the tool parameters. In my proposal, I had loosened this a bit: At least one field in the subject must be obtained from the token.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Agreed it was redundant. Fixed in 083f6b6, with one tweak to your suggestion: I went with the literal `tools/call` rather than `tool/invoke`, since every other default mapping uses the JSON-RPC method name as the action (`tools/list`, `resources/read`, etc). Keeps the action-is-the-method pattern uniform, and the tool name now lives only in `resource.id`. Shout if you feel strongly about `tool/invoke`.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Agree, and this is now spelled out. 083f6b6 rewrites the Architecture section around both deployment shapes: gateway as PEP (it picks up each tool's declared mapping from the `tools/list` response it proxies, no out-of-band config, so the server defines how its operations are authorized and the gateway enforces it, exactly your generic-IT-team point) and server as PEP when there's no gateway. Client visibility is kept as a side benefit for discoverability rather than the headline.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Took this wording, adapted for the envelope rework: 'a template, corresponding to a distinct operation, for an AuthZEN Authorization API request'. In 083f6b6.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Added to the framework terminology in 083f6b6.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Done in 083f6b6, the envelope definitions now cite Section 6 (Access Evaluation) and Section 7 (Access Evaluations) explicitly.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Done in 083f6b6 (Sections 6 and 7).

profiles/authzen-coaz-framework-1_0.md
alexolivier: Done in 083f6b6.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Added in 083f6b6: a non-normative Example Mappings section right after the mapping structure, one `evaluation` example and one `evaluations` example. Followed the SSF Section 5 shape.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Added in 083f6b6: an illustrative HTTP-API default mapping plus a declared mapping that overrides it for one route.

profiles/authzen-coaz-framework-1_0.md
alexolivier: Agree. Renamed the section to 'Denials and Errors' and it now says a denial is the normal course of policy enforcement, not a fault. It stays in the list only because it shares the consequence (operation doesn't proceed). 083f6b6.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Fixed in 083f6b6, swept the whole doc.

profiles/authzen-mcp-profile-1_0.md
alexolivier: This wording changed under the envelope rework: the profile now targets both the Access Evaluation and Access Evaluations APIs, so the abstract keeps the generic 'Authorization API' and the specific API is selected by the mapping envelope.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Done in 083f6b6.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Flipped in 083f6b6, scopes point leads now.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Dropped in 083f6b6.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Done in 083f6b6.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Took your wording in 083f6b6.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Fair, it was cryptic. Now reads: every mapping in this profile (including the defaults) contains CEL expressions, so a conforming PEP has to include a CEL evaluator, there's no expression-free conformance level. 083f6b6.

profiles/authzen-coaz-framework-1_0.md
alexolivier: I get the appeal during development but I don't want the spec to bless it. A fail-open PEP is indistinguishable from no PEP at the worst possible moment (a PDP outage), and dev convenience is easy to get without spec support: point the PEP at a PDP with an allow-all policy and you've got fail-open in practice, with an explicit artifact you have to delete before prod.

So fail-closed stays a MUST in the framework. Happy to add a non-normative note describing the allow-all-policy pattern for dev environments if you think it'd help.

profiles/authzen-coaz-framework-1_0.md
alexolivier: I'm torn on this one. The MCP profile is already unambiguous (CEL, no alternatives), so the flexibility only exists at the framework layer. The reason I kept the contract abstract is future profiles: an OpenAPI or HTTP profile community might reasonably want JSONPath or similar, and if the framework hard-mandates CEL we'd have to rev the framework itself to admit them.

Middle position: the framework names CEL as the default and RECOMMENDED language, a profile MAY override with justification, and the expression contract stays as the thing that makes an override safe. If the WG prefers hard-mandating CEL I won't die on this hill.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Happy to standardise the pattern. One wrinkle: @ogazitt suggested 'binding' for a concrete instantiation in the earlier thread and I'd agreed, and your proposal keeps 'profile' with a dash (COAZ-MCP). The two compose fine honestly: the framework is COAZ, each spec is a profile named COAZ-<protocol>.

Unless there's still a pull toward 'binding' I'll sweep both docs to COAZ / COAZ-MCP. @ogazitt any strong feelings before I do the rename?

profiles/authzen-mcp-profile-1_0.md
alexolivier: Right now all `notifications/*` are pass-through, mostly because they carry no `id`, so there's no JSON-RPC error channel to report a denial on. A server that wants every notification authorized can still do it today as the PEP: evaluate, and silently drop on deny, since no response is expected anyway.

If we want that behaviour to be interoperable rather than implementation-specific, I'd add a per-method opt-in: default mappings for notifications exist but are treated as pass-through unless the deployment enables them. Worth doing for draft 1, or park it?

profiles/authzen-mcp-profile-1_0.md
alexolivier: No error here: `get_local_weather` falls back to the default `tools/call` mapping (subject from the token, action `tools/call`, resource type `tool`, id from the tool name). That's the point of the default mappings: every method gets authorized without any per-tool declaration.

The only hard error is a method with neither a default nor a declared mapping that's also outside the pass-through list, which is about future MCP methods failing closed. Your `coaz:not_required` point is the separate question of whether a tool can opt out of authorization entirely; replying on that thread.

profiles/authzen-mcp-profile-1_0.md
alexolivier: With the default-mapping model the cost of not declaring a mapping is a coarse check, not an error (see my reply on 6.9): the PDP still gets asked 'can this subject call this tool', and a permissive policy makes that a cheap yes.

My hesitation with `coaz:not_required` is that it moves the bypass decision from policy (central, auditable, changeable without redeploying the server) into the tool schema (authored by the thing being authorized). A sloppy or hostile server could mark everything not_required and a gateway PEP would skip enforcement entirely.

If the goal is 'don't make a PDP round-trip for trivially public tools', I'd rather that be a PEP/deployment config decision than a server-asserted flag. Open to it if we scope it as advisory (a PEP MAY honour it), but a MUST-honour flag feels like a hole.

profiles/authzen-mcp-profile-1_0.md
alexolivier: The customer-service bot case is real, I've hit it too. Where I landed for this draft: `subject.id` comes from the token because in the gateway-as-PEP model the declared mapping is authored by the party being authorized. If the mapping can set the subject from a tool parameter, the server can have the PDP evaluate any principal it likes for a caller that never authenticated as them.

The on-behalf-of claim designation (the subject-identity-claim section) covers part of your scenario, when the user id is somewhere in the token. For the email-bot case the user genuinely isn't in the token, and I think the honest modelling is: the authenticated subject IS the agent (or its service identity), and the customer rides along as context or a resource attribute (`context.on_behalf_of` from `$params`), which policy can use without us pretending it's an authenticated identity.

Your 'at least one subject field from the token' rule is an interesting middle ground; my worry is policies rarely check which field was the anchored one. Can we take this one to the WG call?

profiles/authzen-coaz-framework-1_0.md
alexolivier: Follow-up: this is in the spec now as of fb4e3c6. The framework names CEL as the default (SHOULD), a profile MAY designate another language and has to name it explicitly, and CEL moved up to the normative references. I kept it SHOULD rather than MUST so a future profile can override with justification; if the WG wants a hard MUST it's a one-word change.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,717 @@
...
+establishes a single, protocol-neutral model: the inputs of an incoming
+operation are projected, through a declarative mapping, into an AuthZEN
+Authorization API request — an Access Evaluation request for a single decision,
+or an Access Evaluations request for several — expressed using the
+Subject-Action-Resource-Context (SARC) model. Mapping values are either literal constants or expressions evaluated
+against the operation's inputs, enabling both fixed field-to-field mappings and
+dynamic, computed mappings. This document does not define a mapping for any
+specific protocol; it defines the common model and the conformance contract
+that individual COAZ *profiles* fulfil — for example, a profile for the Model
+Context Protocol (MCP), for HTTP APIs, or for OpenAPI-described routes.
```
tulshi: This specification assumes CEL as the language for writing expressions. Specific profiles MAY use other expression languages.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,717 @@
...
+This specification defines COAZ (Compatible with OpenID AuthZEN, pronounced
+"cozy"), a framework for mapping the information model of an arbitrary protocol
+or interface into a request to the OpenID AuthZEN Authorization API. COAZ
+establishes a single, protocol-neutral model: the inputs of an incoming
+operation are projected, through a declarative mapping, into an AuthZEN
+Authorization API request — an Access Evaluation request for a single decision,
+or an Access Evaluations request for several — expressed using the
+Subject-Action-Resource-Context (SARC) model. Mapping values are either literal constants or expressions evaluated
+against the operation's inputs, enabling both fixed field-to-field mappings and
+dynamic, computed mappings. This document does not define a mapping for any
```
tulshi: "document" -> "specification"

tulshi: I see the word "document" being used elsewhere too, it would be good to do a pass for consistency.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,717 @@
...
+- The **input** side is specialized. The set of available input variables, the
+  syntax used to write mappings, the location where a mapping is stored, and
+  the way errors are reported back to the caller all depend on the protocol.
+  These are defined by a COAZ *profile*.
+
+This document defines the output-side model and the conformance contract that
+every profile fulfils. It does not, by itself, define a mapping for any
+protocol. Companion profiles — such as the COAZ Profile for the Model Context
+Protocol {{COAZMCP}} — bind this framework to a concrete information model.
+
```
tulshi, line 144: Expressions in COAZ are, by default, written as CEL expressions. Fields from the operation inputs MAY be used in the CEL expression evaluation.

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,717 @@
...
+Every leaf value in a mapping is either a literal or an expression:
+
+- A **literal** is used verbatim as a constant in the constructed request.
+
+- An **expression** is evaluated, in the profile's expression language, against
+  the operation's input variables, and its result is used in the constructed
+  request.
+
+A profile MUST define an unambiguous discriminator that distinguishes a literal
```
tulshi: Since we are standardizing on CEL (as the default), should we standardize on "$" as the default discriminator, and $$ as the escape? Should we also standardize that the "$" designating an expression must be at the beginning of the string to be considered a discriminator?

profiles/authzen-coaz-framework-1_0.md
```diff
@@ -0,0 +1,717 @@
...
+
+~~~ json
+{
+  "evaluations": {
+    "subject": { "type": "identity", "id": "$token.sub" },
+    "evaluations": [
+      { "action": { "name": "read" },
+        "resource": { "type": "folder", "id": "$request.source" } },
+      { "action": { "name": "write" },
+        "resource": { "type": "folder", "id": "$request.destination" } }
```
tulshi, line 335: Should we add an example where there is an expression, but the expression includes a literal value, e.g.: "$request.destination'&type=destination'"? OR "$'destination='request.destination"

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -124,265 +137,183 @@ when, and only when, they appear in all capitals, as shown here.
...
-: A service that exposes tools, resources, and prompts to MCP clients via
-  the Model Context Protocol.
+: A service that exposes tools, resources, and prompts to MCP clients via the
+  Model Context Protocol.
 
 MCP Gateway:
-: An intermediary that sits between an MCP client and one or more MCP
-  servers. A gateway MAY perform authorization checks on behalf of the
-  MCP server.
+: An intermediary between an MCP client and one or more MCP servers. A gateway
```
tulshi: An *optional* intermediary...

profiles/authzen-mcp-profile-1_0.md
```diff
@@ -124,265 +137,183 @@ when, and only when, they appear in all capitals, as shown here.
...
+an on-behalf-of claim `C`, every use of `$token.sub` as `subject.id` in this
+profile (in default mappings, in declared mappings, and in the verification of
+{{declared-mappings}}) is read as `$token.C`. The agent identity remains in
+`context.agent` (typically `$token.?client_id`) regardless. The designated claim
+MUST be agreed between the PEP and the token issuer; absent any designation, the
+subject-identity claim is `sub`.
+
+# Expressions and Literals {#expressions}
+
+This profile uses Common Expression Language {{CEL}} as its expression language,
```
tulshi: I think we need to commit more strongly to CEL in the framework, and then eliminate this language from the profiles.

profiles/authzen-mcp-profile-1_0.md
ogazitt: I like "binding" better (for the MCP-specific binding of the COAZ framework). But I'm not sure why @tulshi's comment above (which is calling COAZ the "Framework" should change what we call the specific instances (we can still call them "bindings").

profiles/authzen-coaz-framework-1_0.md
alexolivier: Added in 4776dde, with one tweak: the sketched forms (`$request.destination'&type=destination'`) aren't valid CEL, so the example uses CEL's actual concatenation operator: `$'urn:doc:' + request.document_id`. The mixing happens inside the expression language; the mapping layer just sees one expression behind the $.

profiles/authzen-mcp-profile-1_0.md
alexolivier: Consolidation done in 4776dde: the framework now owns the CEL default, the $/$$ discriminator, and the .? absent-value semantics, and COAZ-MCP's expressions section just declares it uses the framework defaults unchanged. One deliberate choice: I kept SHOULD rather than MUST for both the language and the discriminator, so a future binding in an environment that genuinely can't host CEL can override without revving the framework. If the WG prefers a hard MUST, both are one-word changes.

tulshi (approved):

alexolivier:
I've made the proposed changes from the call yesterday. If I could get a final review of that last commit @tulshi @ogazitt  that would be great thanks

README.md
```diff
@@ -7,9 +7,13 @@ The AuthZEN authorization API is versioned in markdown at `api/authorization-api
...
 
-## Draft MCP Profile
+## Draft COAZ Framework
 
-The AuthZEN profile for Model Context Protocol (MCP) Tool Authorization is at `profiles/authzen-mcp-profile-1_0.md`. The HTML version is available [here](https://openid.github.io/authzen/authzen-mcp-profile-1_0.html).
+COAZ (Compatible with OpenID AuthZEN) is a protocol-neutral framework for mapping the information model of an arbitrary protocol or interface into a request to the AuthZEN Authorization API. It is at `profiles/authzen-coaz-framework-1_0.md`. The HTML version is available [here](https://openid.github.io/authzen/authzen-coaz-framework-1_0.html).
+
+## Draft COAZ-MCP Binding
+
+COAZ-MCP is the COAZ binding for the Model Context Protocol (MCP), defining how MCP JSON-RPC messages map into AuthZEN Authorization API requests. It is at `profiles/authzen-mcp-profile-1_0.md`. The HTML version is available [here](https://openid.github.io/authzen/authzen-mcp-profile-1_0.html).
```
ogazitt: I believe these are outdated - should be `profiles/authzen-coaz-mcp-binding-1_0.md` and `https://openid.github.io/authzen/authzen-coaz-mcp-binding-1_0.html`.

ogazitt (approved):
I've reviewed the two new sections (subject.id trust anchoring and removing of opt-out) and they both look good.

I found a slight bug in the README for the COAZ MCP binding pointers - see comment and please fix before merging.

tulshi (approved):

PR #548, opened 2026-07-03
Profile/obligations

baboulebou:
First draft of the AuthZEN Obligations profile.
- Takes into account the latest group discussions (as of 07/03/2026)
- Includes: Obligation definition, normative Obligations and their IANA registration
- Obligation metadata for PDPs
- Obligation negotiation protocol for PEP-PDP interaction.

### DIFF
--- a/profiles/authzen-obligations-profile-1_0.md
+++ b/profiles/authzen-obligations-profile-1_0.md
@@ -10,7 +10,7 @@ workgroup: OpenID AuthZEN
 keyword:
   - authorization
   - obligations
-  - AuthZen
+  - AuthZEN
   - fine-grained authorization
   - policy enforcement
 stand_alone: true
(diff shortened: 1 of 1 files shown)


ogazitt:
@baboulebou This is a really solid first draft, thank you!

I have a few comments:
1. s/AuthZen/AuthZEN/g
2. s/GRANT/PERMIT/g
3. it would be great to have the `obligation` type have a mandatory `type` and `id` as you suggest, but then put any type-specific members in an `ext` JSON object. This makes schema (both for jsonschema and especially for protobuf mappings) much more valuable since the `obligation` doesn't have to be typed as "struct" which doesn't provide any validation.
4. I don't think the search results add a `context` for every result in the array - I believe https://openid.github.io/authzen/#name-the-search-api-response shows that `context` is global for the entire response.  The examples all confirm this. And I think for the purpose of obligations, this is completely fine - having context/obligations per result would be overkill IMO.

For expediency I have approved, subject to the changes I'd like to see before you merge this.

ogazitt (approved):
See my review feedback before merging

profiles/authzen-obligations-profile-1_0.md
```diff
@@ -0,0 +1,562 @@
...
+  "vendor": "example-dlp-suite",
+  "action": "watermark",
+  "watermark_text": "CONFIDENTIAL - jdoe@example.com - 2026-07-03"
+}
+~~~
+{: #fig-obligation-custom title="Non-normative example of a custom obligation"}
+
+# Discovery: PDP Metadata Extension {#metadata-extension}
+
+{{AUTHZEN}} defines a metadata discovery mechanism by which a PEP learns the capabilities of a PDP. This profile adds a new metadata member, `obligations_supported`, an array of Strings, each of which MUST be either an Obligation Type registered per {{iana-obligation-types}}, or the literal value `custom`, that the PDP is capable of issuing.
```
graphel: I know it's an aesthetic preference; but I'd prefer `supported_obligations` as a naturally readable phrase.

graphel: We should also offer support for a wildcard like `*` here; so that the PEP can efficiently indicate support for a broad range of obligation types.

The PEP then fails closed when it receives an obligation type it does not support.

So
```
 { "context" : { "supported_obligations": "vnd.cel.filter.*" } } 
```
or even 
```
 { "context" : { "supported_obligations": "*" } } 
```
for maximum wire-format efficiency.

baboulebou: Not sure I agree with that one. What does it mean "all obligation"? It assumes the PEP "knows" ahead of time all the obligations that the PDP supports, and can enforce all of them. Otherwise, saying * for obligations it can't enforce or support is rather meaningless. Not going to make this change, we can discuss with the group if necessary.

baboulebou: Agreed. Change made.

ogazitt: I agree that “*” is an anti-pattern for capability negotiation. The set of known obligations will grow over time and a PEP that was deployed at time T1 can’t possibly know that it can support all the obligations emitted by a PDP deployed at a later time T2.

mtrimpe: Switching back to my regular account; but a wildcard is quite common in negotiation. See https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2 for a well known example.

In the short term it won't be needed as much but once we get rich PEPs which support dozens of obligations the current pattern of proactively sharing all supported obligation types will add a lot of overhead.

The wildcard can indicate support for obligations in general and a commitment that the PEP will evaluate the type and either understand it and fulfill the obligation; or deny the request.

PR #551, opened 2026-07-09
Add HTML and Text rendering for obligations profile.

baboulebou:


baboulebou:
This branch is messed up, will submit another PR

PR #559, opened 2026-07-16
Add Post Oak Labs (AINumbers) PDP to interop registry

collectrix:
Adds Post Oak Labs' AuthZEN Policy Decision Point to the todo backend interop registry.

- PDP endpoint: https://mcp.ainumbers.co/access/v1/evaluation
- Implements the AuthZEN 1.0 Authorization API evaluation shape (subject/action/resource/context)
- Live and reachable now

Added under both `authorization-api-1_0-01` and `authorization-api-1_0-02` alongside the existing entries.


ogazitt (commented):
Did you run `yarn test` in interop/authzen-todo-backend?  The PDP returns a false decision for every scenario because of at least two requirements that don't exist in the spec but your PDP enforces (there may be others but I'm not broadly testing your PDP - just looking at whether it passes the test suite).

```
curl -X POST -d '{"subject":{"type":"user","id":"CiRmZDI2MTRkMy1jMzlhLTQ3ODEtYjdiZC04Yjk2ZjVhNTEwMGQSBWxvY2Fs"},"action":{"name":"can_delete_todo"},"resource":{"type":"todo","id":"7240d0db-8ff0-41ec-98b2-34a096273b93","properties":{"ownerID":"summer@the-smiths.com"}}}' -H 'content-type: application/json' https://mcp.ainumbers.co/access/v1/evaluation
{"decision":false,"context":{"error":"malformed_request","detail":"context is required"}}
```

Context is *not* required for an AuthZEN request. The test harness which tests this particular scenario does not use context.

Simply being curious, I added a `"context": {}` to the request to see whether that was the only issue. It looks like your PDP requires a custom context.gate property (not in the spec).

```
curl -X POST -d '{"subject":{"type":"user","id":"CiRmZDI2MTRkMy1jMzlhLTQ3ODEtYjdiZC04Yjk2ZjVhNTEwMGQSBWxvY2Fs"},"action":{"name":"can_delete_todo"},"resource":{"type":"todo","id":"7240d0db-8ff0-41ec-98b2-34a096273b93","properties":{"ownerID":"summer@the-smiths.com"}},"context":{}}' -H 'content-type: application/json' https://mcp.ainumbers.co/access/v1/evaluation
{"decision":false,"context":{"error":"malformed_request","detail":"context.gate must be an OCG §21.4 gate: {input, rules[], default}"}}%
```

Stepping back, given that the `interop` directory contains 4 scenarios we used for interop testing as we baked the spec towards v1, it's not the best place to invest your cycles.

I would instead focus on passing the certification test suite. See https://openid.github.io/authzen/authorization-api-1_0-certification-scenario.html

collectrix:
Thanks for the careful review, and you're right on both counts. The PDP was requiring `context` and a bespoke `context.gate`, which is a design mistake on our end, not the spec's; AuthZEN correctly makes `context` optional, and there's no such gate in the standard. It made the PDP return `false` for every conformant request.

I've reworked it: `context`/gate are now optional, and by default the PDP decides from its own server-side policy, and it passes all eight certification fixtures locally (alice/bob read/write, archived + admin, soft/hard delete), with identical decisions with or without `context`. Batch `/evaluations` is in too. Our OCG execution-hash receipt stays as an additive extension on the response, never a requirement for a plain decision.

Taking your steer to focus on the certification suite rather than the interop registry. I'll wire up `authzen-todo-backend` against it and come back through the cert path once it's green. Appreciate the pointer.

collectrix:
Update: the PDP is live at `https://mcp.ainumbers.co/access/v1` and now runs cleanly against the official `authzen-todo-backend` harness, scenarios 00 and 01 pass 40/40, scenario 02 passes 37/40. The three in 02 are the harness's exact-JSON-equality check flagging our additive `execution_hash` receipt field on the batch endpoint, not decision mismatches, the access decisions are correct, confirmed by direct curl against each case. `context` is optional and the batch and search endpoints are in place; the receipt rides as an additive extension, so a plain AuthZEN response is unchanged. Thanks for your feedback, can take the next step on the formal certification path whenever it makes sense.

PR #562, opened 2026-07-17
Rework publish pipeline as a build matrix

alexolivier:
## Problem / Intent

The publish workflow repeats the same convert → render → upload block once per document (8 copies and growing), and the publish job downloads each artifact with its own step. Adding a document means copy-pasting ~25 lines across two jobs, and it's easy to miss one — the COAZ-MCP binding was built but not deployed until #560 added the missing download step.

## Approach

Turn the document list into data: `build-rfc` is now a matrix where each document is a 3-line entry (artifact name, source markdown, output HTML name), sharing a single render/upload step sequence. The main spec's `index.html` copy is a matrix field rather than a dedicated step. Each leg uploads an `rfc-*` artifact whose layout mirrors the published paths, so the publish job collects everything with one pattern download — a new document can never be built-but-not-published again. The archived 0_0 submission is copied straight from the checkout instead of round-tripping through an artifact.

Published output is unchanged: same HTML filenames at the site root, same XML paths, `index.html` still the working draft (verified the artifact layout against the live site's published paths, plus an end-to-end render of the main spec with the same tool versions).

Documents now build in parallel with `fail-fast: false`, so one broken document doesn't mask errors in the others. Note the check names change from `build-rfc` to `build-rfc (<doc>)` — if branch protection pins the old name, it will need updating.

### DIFF
--- a/.github/workflows/jekyll-gh-pages.yml
+++ b/.github/workflows/jekyll-gh-pages.yml
@@ -6,10 +6,44 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
+  group: publish-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   build-rfc:
     runs-on: ubuntu-latest
-
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: authorization-api-1_0
+            source: api/authorization-api-1_0.md
+            html: authorization-api-1_0.html
+            # The working draft of the spec is also published as the site index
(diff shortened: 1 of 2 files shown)


davidjbrossard:
This works for me. We need to document it somewhere though. In the readme.MD? Under `Building the spec`?

davidjbrossard (dismissed):
Update README.md so it contains the new instructions

baboulebou (approved):

PR #567, opened 2026-07-20
Add normative Filter obligation type.

baboulebou:
Added a normative Filter obligation, for forcing a PEP to apply a filter to any PERMIT response or Search result.


davidjbrossard (approved):
I am fine with these changes

PR #611, opened 2026-08-07
Add binding_hash conformance vectors

### LINKED ISSUE #556: Conformance vectors for binding_hash (JCS + SHA-256)
## Summary

Add byte-identical conformance test vectors for the denial-binding `binding_hash` construction, so an independently implemented PDP and Access Request Service have a pass/fail target instead of a prose-only specification.

## Background

`binding-token-integrity` pins the hashed denial-binding form by recomputation over a JCS ([RFC 8785](https://www.rfc-editor.org/rfc/rfc8785)) serialized JSON object plus SHA-256. That is correct, but prose-pinned canonicalization is exactly where independently implemented parties quietly diverge (the field set, nesting, and how `binding_context_members` selects the hashed context members).

Raised in PR #532 review, where a reviewer offered to contribute vectors already validated byte-for-byte across TypeScript, Python, and Go.

## Scope

Vectors for:

- **Single-item `binding_hash`:** `base64url(SHA-256(JCS({"subject": <Subject>, "resource": <Resource>, "action": <Action>, "context": <authz-relevant Context>})))`, with `subject.properties.act` removed.
- **Bulk `binding_hash`:** the `{"subject": <Subject>, "items": [{"resource": <Resource>, "action": <Action>}, ...], "context": <authz-relevant Context>}` form, preserving bound item order.
- **`binding_context_members` selection:** which context members feed the hash (the subtlest part, and the one most worth a vector).
- **Edge cases:** absent vs `null` members, number canonicalization, array ordering, and an empty authorization-relevant context.

## Where

An interop test-vectors location, or a draft appendix. TBD.

## Depends on

Freezing the `binding_hash` field object (currently specified in `binding-token-integrity`); tracks PR #532.

Math1987:
## Summary

- add seven repository-level, non-normative known-answer vectors for the Access Request Approval `binding_hash` construction under `interop/test-vectors/binding-hash`
- cover single-item and bulk inputs, `subject.properties.act` removal, `binding_context_members` selection, absent versus `null`, JCS number serialization, and array/item ordering
- add a dependency-free Node verifier alongside the vectors
- keep the published specification self-contained and unchanged

Closes #556

## Validation

- `node interop/test-vectors/binding-hash/verify-binding-hash-vectors.mjs`
- byte comparison against `canonicalize@3.0.0`
- independent Python comparison of canonical bytes, SHA-256 hex, and unpadded base64url digests
- `git diff --check`

## IPR

The OpenID Foundation Contribution Agreement has been signed and submitted.


Math1987:
Hi @alexolivier,

I’m following up on these binding_hash conformance vectors for the Access Request and Approval Profile. The PR remains limited to informational vectors, a dependency free verifier, and the related documentation link. I would be glad to make any changes needed for review.

alexolivier:
Hey, thanks for the additions. We don't really have a precedent of having these scripts being referenced from a spec before. So I'm not sure how we approach this, given we don't want a published spec to point at something that can move if it just sits in a Git repo.

Math1987:
Thanks, that makes sense. I agree that a published specification should not point to a mutable artifact on the repository main branch.

I updated the PR in ecb1b63 to remove the specification link and the profile-level Make target, and moved the vectors and verifier to `interop/test-vectors/binding-hash`. The published profile is now unchanged; the corpus is explicitly repository-level, non-normative interoperability material.

I also saw the more extensive corpus discussed in #556. I am happy to align this PR with that work rather than introduce competing vector sets once the preferred location and scope are confirmed.

Does this repository-level interop location, with no reference from the published specification, address your concern?

alexolivier (approved):

PR #626, opened 2026-08-23
COAZ-MCP: bind permits to evaluated operations

### LINKED ISSUE #603: [COAZ-MCP] Bind a permit to the MCP operation actually forwarded
## Problem

The current COAZ-MCP processing algorithm populates the mapping inputs from an
MCP request, obtains a permit/deny decision, and then allows the message to
proceed. It does not explicitly state that the operation forwarded or executed
must still be the operation that produced the authorization request.

For example, a gateway pipeline could evaluate:

```json
{
  "method": "tools/call",
  "params": {
    "name": "payments.send",
    "arguments": { "amount": 500 }
  }
}
```

and obtain a permit, after which a later middleware stage changes `amount`,
`name`, or even `method` before forwarding. The PDP response is valid, but the
PEP has applied it to a different operation.

This is separate from obligations/constraints in #521 and from post-action
receipts. It is the pre-execution authorization-to-use boundary.

## Suggested clarification

In the Processing Algorithm and/or Mapping Integrity security consideration:

- The PEP MUST evaluate the final normalized operation that it will forward or
  execute.
- A permit applies only while the inputs that selected the mapping or
  contributed to the constructed AuthZEN request remain semantically unchanged.
- If any such input changes after evaluation, the PEP MUST re-evaluate the final
  operation or refuse it; it MUST NOT reuse the earlier permit.
- Rewrites or other known transformations that can affect mapping selection or
  mapped values MUST occur before mapping resolution and evaluation.
- If the MCP server re-authorizes a gateway-forwarded request, the same rule
  applies between the server's evaluation and local execution.

“Semantically unchanged” should be defined over mapping-selection inputs and the
binding's input-variable values, not over raw JSON bytes, so harmless
serialization or transport normalization remains possible. Fields deliberately
excluded by a medium-grained mapping are not implicitly made fine-grained by
this rule.

## Candidate negative vectors

1. unchanged `tools/call` after permit → forward;
2. mapped tool name changed after permit → re-evaluate or refuse;
3. mapped arguments changed after permit → re-evaluate or refuse;
4. method changed after permit → re-evaluate or refuse;
5. mapped trusted context changed after permit → re-evaluate or refuse.

Would the WG prefer this as an additional normative processing step, or as a
Mapping Integrity requirement plus conformance vectors?

FutureEnterprises:
## Summary

- apply known rewrites and normalizations before mapping selection and evaluation
- require the PEP to re-evaluate or refuse when the method, selected mapping, or binding input values change before forwarding
- state explicitly that an AuthZEN decision does not cover source inputs omitted by the selected mapping

Closes #603.

## Why

Issue #603 identifies a time-of-check to time-of-use gap between the operation evaluated by the PDP and the operation ultimately forwarded or executed.

The added authorization-granularity text records a separate boundary found while reproducing that issue. A valid mapping can intentionally omit an input. Two source operations that differ only in that input can then construct the same AuthZEN request and receive the same decision. That is acceptable for coarse-grained authorization, but the omitted input was not evaluated by the PDP and must not be represented as if it were.

## Validation

- `kramdown-rfc2629` render passed
- `xml2rfc` HTML render passed
- `git diff --check` passed
- source-pinned external reproduction: 9 of 9 cases passed
- Vitest reproduction: 7 of 7 tests passed

Reproduction artifact:
https://github.com/emiliaprotocol/emilia-protocol/blob/feat/authzen-coaz-mapping-integrity-20260823/conformance/composition/coaz-translation-v0.1/AUTHZEN-CONTRIBUTION.md

## Claim boundary

The reproduction uses a mapping and toy PDP created for the corpus. It does not claim that a deployed AuthZEN, COAZ, MCP, gateway, or PDP product is vulnerable. CAID is demonstrated only as an optional relying-party content-correlation mechanism and is not a normative dependency of this patch.

## IPR

The OpenID Foundation Contribution Agreement for EMILIA Protocol, Inc. and the AuthZEN Working Group was completed and countersigned on August 21, 2026.

### DIFF
--- a/profiles/authzen-coaz-mcp-binding-1_0.md
+++ b/profiles/authzen-coaz-mcp-binding-1_0.md
@@ -792,21 +792,25 @@ so on the basis of the verified `subject.id`, not the declared `subject.type`:
 
 When an MCP message is processed, the PEP MUST:
 
-1. Determine the JSON-RPC `method`. If it is a pass-through operation
+1. Apply every rewrite or normalization known to the PEP that can change the
+   JSON-RPC `method`, mapping selection, or a value exposed through a binding
+   input variable. The resulting message is the evaluated message.
+
+2. Determine the JSON-RPC `method`. If it is a pass-through operation
    ({{default-mappings}}), allow it without calling the PDP.
 
-2. Select the applicable mapping: for a `tools/call` whose tool declares an
+3. Select the applicable mapping: for a `tools/call` whose tool declares an
    `x-authzen-mapping`, use the declared mapping; otherwise use the default mapping
    for the method. A method that is neither in the pass-through set nor has a
    mapping MUST be denied ({{default-mappings}}).
 
-3. Populate `params` from the request's `params` object and `token` from the
+4. Populate `params` from the request's `params` object and `token` from the
    decoded, validated access token claims.
 
-4. Resolve the mapping: literals verbatim, `$`-prefixed values by evaluating the
+5. Resolve the mapping: literals verbatim, `$`-prefixed values by evaluating the
    CEL expression ({{expressions}}).
 
-5. Anchor the subject identity: where a mapping sets `subject.id` to the
+6. Anchor the subject identity: where a mapping sets `subject.id` to the
    subject-identity claim ({{subject-identity-claim}}) — as every default mapping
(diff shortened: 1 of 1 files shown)


alexolivier (approved):
This is a great addition - thank you!

PR #637, opened 2026-09-04
Token issuance: fix the action name bounds, plus an editorial pass

ogazitt:
One substantive fix and an editorial pass over the issuance profile, from
Dave Hyland's review of the adopted text.

The action name constraints created ambiguity, so the draft now limits short
names to `[a-z][a-z0-9_]{0,20}`, which means a composed action name
(`issue:<token-type>:<grant-type>`) will not be longer than 49 characters.
No initial registration is affected: the longest entries in either vocabulary
are `authorization_code` and `client_credentials` at 18 characters.

The rest is prose, in four places:

- §1.1 Design Goals drops a paragraph that defended the five-tuple rule
  against an objection nobody had raised, and starts on the claim itself.
- §4.1 Colocation and Latency goes from three paragraphs to two sentences.
  The claim that the profile "implies no additional lookup" was about the AS
  assembling the request and read as though it were about the cost of
  evaluating the policy, which is not something this document can promise.
- §4.3 The Seam Inside the Authorization Server says what it means: the
  values are spread across one subsystem, and finding a single point that
  holds them all is the work of adopting the profile.
- §5.3.1 glosses the gate action examples, since
  `issue:access_token:authorization_code` does not obviously read as a token
  type followed by a grant type.

Builds clean; the only xml2rfc warning is the usual missing `submissionType`.


alexolivier (approved):

davidjbrossard (approved):
Looks good to me too

PR #650, opened 2026-09-10
Specify response shape when evaluations array is absent or empty

### LINKED ISSUE #639: Response shape is unspecified when an Access Evaluations request omits `evaluations`
Section 7.1 says:

> If an `evaluations` array is NOT present or is empty, the Access Evaluations Request behaves in a backwards-compatible manner with the (single) Access Evaluation API Request ({{access-evaluation-request}}).

This is scoped to the request. Neither 7.1 nor 7.2 states what the response looks like in that case. The only hint is in 7.2:

> In case the `evaluations` array is present, it is RECOMMENDED that the `decision` key of the response be omitted.

which conditions the omission on the array being present, implying `decision` is present when it is absent, but never states that as a requirement.

The certification scenario does state it. `certification/authorization-api-1_0-scenario.md` c-3-4-2 and c-3-4-3 require the PDP to "return an Access Evaluation response", give `{"decision": true}` as the expected body, and validate the response structure.

For the empty-array case the plural-envelope reading is excluded outright. c-3-3-1 says:

> The `evaluations` array in the response MUST have the same number of elements as the `evaluations` array in the request.

A PDP that answers `"evaluations": []` with a one-element array violates that MUST regardless of how c-3-4-3 is read. Nothing in the API document signals this.

Suggested wording for 7.1:

> If an `evaluations` array is NOT present or is empty, the Access Evaluations Request behaves in a backwards-compatible manner with the (single) Access Evaluation API Request ({{access-evaluation-request}}), and the PDP responds with an Access Evaluation Response ({{access-evaluation-response}}) rather than an Access Evaluations Response.

alexolivier:
Fixes #639

Section 7.1 says an Access Evaluations Request with no `evaluations` array (or an empty one) behaves like a single Access Evaluation Request, but neither 7.1 nor 7.2 states what the response looks like. The only hint is the RECOMMENDED omission of `decision` in 7.2, which is conditioned on the array being present.

This adds one clause to the 7.1 sentence: the PDP responds with an Access Evaluation Response rather than an Access Evaluations Response. This matches what the certification scenario already requires in c-3-4-2 and c-3-4-3, and is consistent with c-3-3-1 (response array length equals request array length), which excludes a one-element `evaluations` envelope for the empty-array case.

### DIFF
--- a/api/authorization-api-1_0.md
+++ b/api/authorization-api-1_0.md
@@ -470,7 +470,7 @@ The Access Evaluation API Request builds on the information model presented in {
 
 To send multiple access evaluation requests in a single message, the PEP MAY add an `evaluations` key to the request. The `evaluations` key is an array which contains a list of objects, each typed as the object as defined in the Access Evaluation Request ({{access-evaluation-request}}), and specifying a discrete request.
 
-If an `evaluations` array is NOT present or is empty, the Access Evaluations Request behaves in a backwards-compatible manner with the (single) Access Evaluation API Request ({{access-evaluation-request}}).
+If an `evaluations` array is NOT present or is empty, the Access Evaluations Request behaves in a backwards-compatible manner with the (single) Access Evaluation API Request ({{access-evaluation-request}}), and the PDP responds with an Access Evaluation Response ({{access-evaluation-response}}) rather than an Access Evaluations Response ({{access-evaluations-response}}).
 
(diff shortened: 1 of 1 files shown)


davidjbrossard:
I am fine with this rewording.

davidjbrossard (approved):

Issue #486, opened 2026-04-14
New fields should be added to the _meta field


tulshi:
From @maxwellgerber:
MCP extensions like UI hang their values off of the _meta field, not on top-level objects - https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx - I know there is a tool annotations IG that is taking a look at this but unsure what the current recommendation is

alexolivier:
The [2026-07-28 spec](https://modelcontextprotocol.io/specification/2026-07-28/basic#_meta) says implementations SHOULD use reverse DNS notation in the `_meta` key, so it would be something like `net.openid.authzen/coaz-mcp`,

So the `tools/list` example from the binding becomes:

```json
{
  "tools": [
    {
      "name": "get_customer",
      "description": "Get customer details",
      "inputSchema": {
        "type": "object",
        "properties": {
          "id":   { "type": "string", "description": "The customer identifier" },
          "case": { "type": "string", "description": "The case being worked on" }
        },
        "required": ["id"]
      },
      "_meta": {
        "net.openid.authzen/coaz-mcp": {
          "evaluation": {
            "subject":  { "type": "identity", "id": "$token.sub" },
            "action":   { "name": "get_customer" },
            "resource": { "type": "customer", "id": "$params.arguments.id" },
            "context":  { "case": "$params.arguments.case" }
          }
        }
      }
    }
  ]
}
```

Two reasons this is better than where we are today. `inputSchema` is a JSON Schema 2020-12 document and the 2026-07-28 spec is explicit that any standard keyword may appear there, so `x-authzen-mapping` is a foreign keyword sitting in a namespace that is not ours. And `_meta` hangs off the `Tool` object itself, which is where the rest of the ecosystem puts this sort of thing - MCP Apps does the same for its UI metadata.

baboulebou:
@alexolivier @tulshi is there a PR with this change ☝️  ?

PR #654, opened 2026-09-17
Add client usage example

### LINKED ISSUE #481: Client usage example
from @maxwellgerber:
Can you provide an example of when this mapping would be useful to reveal to the Client, where the equivalent information wouldn't be part of the existing inputSchema descriptions? Unless the client understands the entire policy orchestration system, I'm not sure that knowing the request mapping  is actually that useful. And I know people have concerns about revealing too much about system authz mechanics to agents out of fear that the agent will poke at things it shouldn't. (Oh, I can't make a request in $USD, let me try $CND and see if that goes through.... etc)

tulshi:
Closes #481


tulshi:
@maxwellgerber. Please review this to check if it addresses your question (Issue #481)

alexolivier:
I think the profile needs to change based on https://github.com/openid/authzen/issues/486 where `_meta` is defined int the spec https://modelcontextprotocol.io/specification/2026-07-28/basic#_meta

maxwellgerber:
Can we find a stronger example? If the tool always requires a `report_id` to function, then that field should be marked as required. That can be done in native JSONSchema - and a client would know it _has_ to supply that field regardless. No need for the client to be aware of `x-authzen-mapping`.

baboulebou (approved):

Issue #638, opened 2026-09-04
Separate gate and scope tuples by resource.type instead of the reserved issue: prefix


ogazitt:
§5.2 requires `resource.type` to be `audience` for both gate tuples and scope tuples, so the only thing distinguishing them is the shape of `action.name`: a gate action matches `issue:<token-type>:<grant-type>`, a scope action is an arbitrary client-chosen scope value. §5.3.1 protects that by reserving the `issue:` prefix. AS-controlled and client-controlled strings end up in one flat namespace disambiguated by pattern matching, and applications lose the ability to choose scope values freely.

§5.3.4 already solves this a different way for authorization detail tuples, and §9.4 says it outright: "`resource.type` is what separates them." Scope tuples are the one tuple kind that doesn't get that treatment.

Proposal: gate tuples keep `resource.type` = `audience`, which is what RFC 8707 and RFC 8693 call it and what lands in `aud`. Scope tuples take a second registered type, `protected_resource`, naming the role the target plays in the sense of RFC 9728. The `issue:` reservation in §5.3.1 can then be deleted, since a client scope value at `protected_resource` cannot collide with a gate action at `audience` however it is spelled.

Touches §5.2 and its one-type rationale, §5.3.1, §5.3.3, §5.3.4, the Token Issuance Entity Types registry, and every example with a scope tuple.

Raised by Roland Baum on the OAuth WG mailing list.

ogazitt:
Resolved by #677, which takes the proposal as filed.

Scope tuples take `protected_resource`; gate tuples keep `audience`. One target
appears under both types in a batch, answering a different question under each.
The `issue:` reservation is deleted, so the profile reserves no scope values
and a deployment keeps its whole scope namespace.

Beyond the sections listed above, two consequences worth noting. A batch now
mixes resource types, so the top-level `resource` can serve as the default for
at most one kind and §5.5.1 says so. And an Action Search at an `audience`
resource returns gate actions only, which §7 previously had to describe as
filtering gates out of a mixed result.

Both companion documents needed changes too: the token exchange binding
restated the rule as `audience` for both tuple kinds, and both companions carry
scope tuples in worked examples.

Thanks Roland.

PR #677, opened 2026-09-23
OAuth issuance: separate gate and scope tuples by resource type

ogazitt:
Closes #638.

Gate tuples and scope tuples both carried `resource.type` of `audience`, so the
only thing telling them apart was the shape of `action.name`, and §5.3.1
protected that by reserving the `issue:` prefix against scope values.
AS-composed and client-chosen strings shared one flat namespace and were
disambiguated by pattern matching.

Scope tuples now take a second registered type, `protected_resource`, naming
the role the target plays in the sense of RFC 9728. Gate tuples keep
`audience`, which is what RFC 8707 and RFC 8693 call the target and what lands
in `aud`. One target appears under both types in a single batch, answering a
different question under each.

The `issue:` reservation is deleted. The profile now reserves no scope values
and a deployment keeps its whole scope namespace.

## Changes

**Token issuance**

- §5.2: `resource.type` is `audience` for a gate tuple and `protected_resource`
  for a scope tuple. The one-type rationale is rewritten; both types still name
  a protocol role rather than a guess at the target's nature.
- §5.3.1: the `issue:` reservation is gone.
- §5.3.3: a scope tuple's resource is the target under `protected_resource`.
- §5.3.4: the gate-versus-authorization-detail paragraph becomes the general
  statement that each of the three tuple kinds is identified by its resource
  type.
- §5.5.1: a batch mixes resource types, so the top-level `resource` can serve
  as the default for at most one kind and the rest carry their own.
- §7: an Action Search at an `audience` resource now returns gate actions only.
- Entity Types registry: `protected_resource` added.
- §9.1 and §9.2 carry scope tuples and gain explicit resources. RFC 9728 added
  to informative references.

**Companions**

- Token exchange §5.3 restates the rule for the tuples it forms; its worked
  example gains a resource on the scope tuple. The two-level evaluation of
  §5.3.1 needs no change, since it constrains `resource.id` and not the type.
- Authorization claims: the worked example gains resources on its scope tuples.

All three documents build clean and stay within the 72-column cap.

Stacked on #676, which is stacked on #675, #673 and #671. Merge in that order.


ogazitt:
Adds a second commit acknowledging Roland Baum and Dave Hyland in the issuance
profile, naming what each review changed. Only that document: the companions'
Acknowledgments sections name no individuals, and both reviews were of the
issuance profile.

